PR #15, opened 2022-06-10
[patch] 9469 adding sls catalog source and channel to update SLS vers…

padmankosalaram:
…ion to use dev image in saas FVT pipeline


sachinjha (approved):
Looks good.

PR #57, opened 2022-08-18
[patch] Include dependencies in Apps when clause

alequint:
Modify when clause of suite_app_install and suite_app_config of applications that depend on other applications. Tasks can only be executed if applications the task depend on are also meant to be installed. We want to avoid problems like this: https://ibm-watson-iot.slack.com/archives/C031Q7W21FZ/p1660753420691909?thread_ts=1660753297.143349&cid=C031Q7W21FZ

### DIFF
--- a/tekton/pipelines/install.yaml
+++ b/tekton/pipelines/install.yaml
@@ -1311,9 +1311,12 @@ spec:
       taskRef:
         name: mas-devops-suite-app-install
         kind: Task
-      # Only install Monitor if a channel has been chosen
+      # Only install Monitor if a channel has been chosen for both Monitor and IoT
       when:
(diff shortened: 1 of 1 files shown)


durera:
This is a desired scenario to support ... e.g. use this pipeline to install (add) a new application to an existing install.  When using this pipeline from a clean start it's the responisibility of the caller to ensure the right combination off applications are used (e.g. the code in the `image/cli/bin/mas` is doing this).  The pipeline is the engine and needs to offer lots of scenarios, otherwise it cannot be multi-purpose/we will end up having to maintain lots of different pipelines.

In our own test environments, at some point we will start testing "does the application do what it's supposed to when an app it integrates with is missing", and this will also be key capability then.

PR #68, opened 2022-09-29
[patch] Add cpd_product_version in Task

alequint:


andrercm (commented):
cool, yeah i think that's all we need to make cli flexible to choose which cpd version we want to install !

PR #74, opened 2022-10-21
[patch] Missing parameter in install pipeline for cp4d

alequint:
Parameter has been added in task but not in pipeline, got the issue when using CLI to install some environment with CPD 4.5.

### DIFF
--- a/tekton/pipelines/install.yaml
+++ b/tekton/pipelines/install.yaml
@@ -857,8 +857,11 @@ spec:
         - name: ibm_entitlement_key
           value: $(params.ibm_entitlement_key)
 
+        - name: cpd_product_version
+          value: $(params.cpd_product_version)
         - name: mas_instance_id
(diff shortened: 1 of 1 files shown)


tekton/pipelines/install.yaml
```diff
@@ -857,8 +857,11 @@ spec:
         - name: ibm_entitlement_key
           value: $(params.ibm_entitlement_key)
 
+        - name: cpd_product_version
```
durera, line 860: Doesn't there need to be some related changes in mascli/functions/xxx, or were those already done and this was just forgotten about?

alequint: Let me move this to draft again... I will double check. I did that PR while working in other changes in devops/ansible-fvt, but did not check the existing flow in CLI... I will, before submitting to review again :+1:

durera:
Already addressed elsewhere, closing PR

PR #106, opened 2022-12-06
[patch] Support Db2 resource customization

sbudhira:
This PR is for code changes made for Single Node OpenShift.

1. Add changes to display Manage app and MAS 8.9 only for SNO.
2. Prompt user for DB2 size parameters and apply to the DB2 install.
3. Add a separate storage class for SNO to prompt only RWO class. SNO uses RWO class.

Tested on bare metal env. Installed LVM storage class and used this code to install MAS and Manage.

https://console-openshift-console.apps.sno4.sarika.donatelli.click/k8s/ns/mas-sno-pipelines/tekton.dev~v1beta1~PipelineRun/sno-install-mtz5p

xZZEm-UDni9-txrYM-f6L7A


image/cli/mascli/functions/pipeline_show_config
```diff
@@ -112,6 +112,15 @@ function pipeline_show_config() {
     echo_reset_dim "ReadWriteMany ............ ${COLOR_MAGENTA}${STORAGE_CLASS_RWX}"
   fi
 
+  reset_colors
+  echo "${TEXT_DIM}"
+  echo_h2 "DB2 Configuration" 
+  echo_reset_dim "DB2_META_STORAGE_SIZE ............ ${COLOR_MAGENTA}${DB2_META_STORAGE_SIZE}"
+  echo_reset_dim "DB2_DATA_STORAGE_SIZE ............ ${COLOR_MAGENTA}${DB2_DATA_STORAGE_SIZE}"
+  echo_reset_dim "DB2_BACKUP_STORAGE_SIZE ............ ${COLOR_MAGENTA}${DB2_BACKUP_STORAGE_SIZE}"
```
durera: This isn't aligned with the rest ...

tekton/pipelines/install.yaml
```diff
@@ -95,6 +95,28 @@ spec:
       description: Optional configuration for mongodb cpu limits
       default: ""
    
+   # DB2 size parameters
```
durera: See the comment in the file:

```
# Don't edit install.yaml directly -- modify install.yaml.j2, and the content in parts/, then run "ansible-playbook generate-install.yaml"
```

Need to update `install.yaml.j2` and  `taskdefs/dependencies/db2.yml.j2` and use the playbook to build this file .. otherwise next time someone does this these changes will be lost.

image/cli/mascli/templates/pipelinerun.yaml
```diff
@@ -34,6 +34,16 @@ spec:
       value: '$DB2_CPU_REQUESTS'
     - name: db2_cpu_limits
       value: '$DB2_CPU_LIMITS'
+    - name: db2_meta_storage_size
```
durera, line 38: Where do these new values get set in non-SNO mode?  `config_pipeline_storage_db2_size` only seems to do anything whern SNO mode == true.

durera (changes requested):

image/cli/mascli/functions/pipeline_show_config
sbudhira: fixed alignment,

image/cli/mascli/templates/pipelinerun.yaml
sbudhira: This is added for SNO only. In non-SNO mode, these will be null and playbook default values will be used.

tekton/pipelines/install.yaml
sbudhira: Fixed

image/cli/mascli/functions/pipeline_config_applications
```diff
@@ -195,42 +199,51 @@ function config_pipeline_applications() {
...
+        if prompt_for_confirm "Install Health & Predict - Utilities"; then
+          channel_select_hputilities || exit 1
+        fi
       fi
     fi
 
-    # Assist
-    if prompt_for_confirm "Install Assist"; then
-      channel_select_assist || exit 1
+    if [[ "$SNO_MODE" != "true" ]]; then
```
alequint: @sbudhira can't we have a single `if [[ "$SNO_MODE" != "true" ]];` statement and add the prompts inside it instead of multiple if statements like this? Any particular reason to be done that way?

image/cli/mascli/functions/pipeline_config_db2
```diff
@@ -0,0 +1,22 @@
...
+        echo
+        echo_h2 "6b. DB2 Storage size, in Gi(Gigabytes)"
+        prompt_for_input "Size of data persistent volume" DB2_DATA_STORAGE_SIZE && export DB2_DATA_STORAGE_SIZE
+        prompt_for_input 'Size of temporary persistent volume' DB2_TEMP_STORAGE_SIZE && export DB2_TEMP_STORAGE_SIZE
+        prompt_for_input "Size of metadata persistent volume" DB2_META_STORAGE_SIZE && export DB2_META_STORAGE_SIZE   
+        prompt_for_input 'Size of transaction logs persistent volume' DB2_LOGS_STORAGE_SIZE && export DB2_LOGS_STORAGE_SIZE
+        prompt_for_input "Size of backup persistent volume" DB2_BACKUP_STORAGE_SIZE && export DB2_BACKUP_STORAGE_SIZE  
+
+    fi
+}
```
alequint: Need an extra line at the end of the file to avoid `lint` issues

image/cli/mascli/functions/pipeline_config_storage_classes_sno
```diff
@@ -0,0 +1,75 @@
...
+# - uds_storage_class [ibmc-block-bronze, ocs-storagecluster-ceph-rbd, managed-premium] (doesn't really need to be bronze)
+
+# Unknown/TBC
+# - appconnect_storage_class - doesn't support auto-select storage classes :(
+# - cpd_metadata_storage_class [ibmc-block-gold, ocs-storagecluster-ceph-rbd, managed-premium]
+# - cpd_primary_storage_class [ibmc-file-gold-gid, ocs-storagecluster-cephfs, azurefiles-premium]
+# - cpd_service_storage_class [ibmc-file-gold-gid, ocs-storagecluster-cephfs, azurefiles-premium] / WD: [ibmc-block-gold, ocs-storagecluster-ceph-rbd, managed-premium]
+
+function config_pipeline_storage_classes_sno() {
+  if [[ "$SNO_MODE" == "true" ]]; then
```
alequint, line 27: do we need to check it again? this function is called when needs to be called (inside an `if` statement like this)... can we remove that and simplify the code?

alequint (changes requested):

image/cli/mascli/functions/pipeline_config_applications
sbudhira: Fixed

image/cli/mascli/functions/pipeline_config_db2
sbudhira: Fixed

image/cli/mascli/functions/pipeline_config_storage_classes_sno
sbudhira: yes, it is needed. The check is only in this file.

alequint: I am asking because pipeline_config_storage_classes_sno is called here https://github.com/ibm-mas/cli/pull/106/files#diff-bccb9816831bc1cb7fcffb7c9f3e1d3f893cc1c88ea4e613517eada55c980a1bR130, which is already inside the if statement like the one you are adding here. So, in the end, the file is called when needed and what is inside it don't need to be rechecked... but that is fine, this is a minor point

alequint (approved):

alequint:
Looks good to me, all the requested changes have been addressed and tests have been performed.

PR #130, opened 2022-12-30
[minor] Add Upgrade FVT Pipeline and Tekton Build System

alequint:
## Overview

This PR contains upgrade-test pipeline as well as automation code to generate and release pipelineruns in artifactory (and some other misc fixes like yml x yaml standards). Other PRs related to this one:

- `ansible-devops` (**merge before this**):
  - Add skip_compatibility_check for upgrade error path test (https://github.com/ibm-mas/ansible-devops/pull/595), originally created by @liyongc
  - Fix timeout in OperatorCondition Checking (https://github.com/ibm-mas/ansible-devops/pull/631) created by @alequint 

- `upg-stable` (as reference only): here is the travis that will trigger upgrade-template: https://github.ibm.com/maximoappsuite/upg-stable/blob/master/.travis.yml. Here we have:
  - minimum variables needed to install, test and upgrade
  - the variables representing values related to upgrade process itself are defined using `source_` and `target_` prefixes. Source and Target of some of them are equal today like ocp_version, cpd_product_version, mas_catalog_version, but are prepared to be different as we evolve upgrade test and script itself. Today, the only source/target vars are the channels of core and applications, following phase 1 of upgrade test epic.
  - difference from regular fvts is that mas_tekton's run-pipeline is used to trigger upgrade-test pipeline, as it uses automated generated pipelunerun released in artifactory, as will be explained along the PR

## Content

- New `upgrade-test` pipeline, which is an integration of 3 other existing pipelines:
  - `install`, `fvt` (which is same as current fvt-core, but plan is expand it to contain fvt of all the applications) and `upgrade`
- Three new Tasks capable to run these three pipelines (in `tekton/task-templates`):
  - `fvt-run-pipeline`
  - `install-run-pipeline`
  - `upgrade-run-pipeline`
  - **Note**: these tasks run a role inside [mas_tekton](https://github.ibm.com/maximoappsuite/mas-tekton) called `run-pipeline`. This role simply download a pipelinerun **template** from artifactory (as released by cli) and apply it in the cluster. It is expected environment is filled with the right variables to be used to fill up pipelinerun template; these variables are set by the run-pipeline tasks previously mentioned (which come from upgrade-test pipeline's taskdefs)
- Added more automation to generate tekton. Considering we have now tasks capable to run pipelines, we want that taskdefs, tasks and pipelineruns being automatically generated. It means that we just need to add parameter in the proper param template file and don't worry with the rest, as the automation will organize it all (add it in the task, in pipeline, in taskdefs, in pipelinerun itself etc). Manual settings are excluded from automation by using a special manualsettings file, so they are excluded from the loop and , must be set manually (it happens when the value does not come from owner pipeline, but we know the value to use). Tekton tree is like this now:

```
> tekton
  > generate-tekton.yml (playbook)
  > pipeline-templates
    > params
      > common.yml.j2
      > upgrade-test.yml.j2
      > upgrade.yml.j2
      > (others already there)
    > upgrade-test.yml.j2
  > pipelinerun-templates
    > fvt.yml.j2
    > install-with-fvt.yml.j2
    > install.yml.j2
    > upgrade-test.yml.j2
    > upgrade.yml.j2
  > task-templates
  > pipelines
  > pipelineruns
  > tasks
```
- Build process has been updated to upload pipelineruns to artifactory same way ibm-mas-tekton.yml. run-pipeline role then will down load the pipelinerun to apply and will have, in their environment all the variables needed to update it.
- Standardize use of `yml` instead of `yaml`. We had a mix of these two extensions in cli; then I picked up `yml` as it seems the one mostly used around, including `ansible-devops` (though even there we have a mix of standards in some parts)
- **Important** to say that in this first drop I will include nothing beyond core tests in fvt pipeline. I want to be sure we keep running that upgrade test against mas core first and then start including the applications one by one.

## Test Evidences

Complete set of pipelines triggered by upgrade-test pipeline. This test has been made with **mas core** and **optimizer**.

image


alequint:
Moved PR back to Draft to make automation for manually set parameters betters... will start/finish work in Jan 18th and hopefully put it in review again.

.github/workflows/build-cli.yml
```diff
@@ -29,16 +29,27 @@ jobs:
       - name: Upload the Tekton definitions
         uses: actions/upload-artifact@v2
         with:
-          name: ibm-mas-tekton-${{ env.VERSION }}.yaml
-          path: ${{ github.workspace }}/image/cli/mascli/templates/ibm-mas-tekton.yaml
+          name: ibm-mas-tekton-${{ env.VERSION }}.yml
+          path: ${{ github.workspace }}/image/cli/mascli/templates/ibm-mas-tekton.yml
```
durera: Wouldn't change this, turns this into a breaking change.  we use `yml` inside ansible code, and `yaml` for things that a user will `oc apply -f` to align with what seems or seemed at the time, the generally accepted standard in both contexts.

Makefile
```diff
@@ -24,10 +24,10 @@ run:
 
 clean:
 	rm image/cli/install-ansible/ibm-mas_devops.tar.gz
-	rm image/cli/bin/templates/ibm-mas-tekton.yaml
+	rm image/cli/bin/templates/ibm-mas-tekton.yml
 
 create:
-	oc apply -f tmp/deployment.yaml
+	oc apply -f tmp/deployment.yml
```
durera: As above, yaml (openshift/kubernetes) versus yml (ansible), don't think we really want to change this.

tekton/pipeline-templates/params/common.yml.j2
```diff
@@ -0,0 +1,81 @@
+# CLI Version
```
durera: Bad merge possibly?  `cli_version` is no more after the last round updates to optimize development flow.

tekton/pipeline-templates/params/common.yml.j2
```diff
@@ -0,0 +1,81 @@
...
+- name: cli_version
+  type: string
+  description: version of mas cli to be used in tasks and dependencies
+  default: ""
+
+# Registry
```
durera: `common` to me implies there are params used in every pipeline, but these `fvt_` params would only be used in the fvt pipelines, not the "normal" ones ... am I misunderstanding the context of "common" here?

tekton/pipeline-templates/params/fvt.yml.j2
```diff
@@ -1,13 +1,3 @@
-# FVT Image Container Registry
```
durera, line 1: See previous comment, I'm confused why these FVT specific params are being moved into common, that doesn't make sense to me.

tekton/pipeline-templates/params/helper/install-list.yml.j2
```diff
@@ -0,0 +1,2 @@
+params:
```
durera: This file seems like an unnecessary layer of abstraction compared to whatever is including this including this content directly .. but might make sense as I get further through the changes ...

tekton/pipeline-templates/params/helper/upgrade-list.yml.j2
```diff
@@ -0,0 +1,2 @@
+params:
```
durera: This file seems like an unnecessary layer of abstraction compared to whatever is including this including this content directly .. but might make sense as I get further through the changes ...

tekton/tasks/cert-manager.yml
```diff
@@ -48,6 +56,11 @@ spec:
         value: $(params.devops_suite_name)
       - name: DEVOPS_BUILD_NUMBER
         value: $(params.devops_build_number)
+      # Optional parameters to categoriz result to be persisted
```
durera: Small typo in various places to find/replace fix `categoriz` -> `categorize`

durera (changes requested):
See individual comments ... 

Also, local built version of `ibm-mas_devops.tar.gz` needs to be deleted before this can be merged.  If there's a branch of ansible-devops that needs to be merged before this put a link in the PR description, e.g. like in this PR https://github.com/ibm-mas/cli/pull/124, and we'll get that released before this goes to master

.github/workflows/build-cli.yml
alequint: The two PRs from ansible-devops mentioned in the description of the PR must be merged first :+1:. I will emphasize they need to be merged. I will remove the zip, goo catch!

alequint:
@durera the two PRs from ansible-devops mentioned in the description of the PR must be merged first. Just emphasized that in the description. I have also removed the ansible-devops zip file form this PR (good catch!)

tekton/pipeline-templates/params/helper/install-list.yml.j2
alequint: https://ibm-watson-iot.slack.com/archives/C0195MVCEUD/p1675098979911869?thread_ts=1675088015.554689&cid=C0195MVCEUD

tekton/pipeline-templates/params/helper/upgrade-list.yml.j2
alequint: https://ibm-watson-iot.slack.com/archives/C0195MVCEUD/p1675098979911869?thread_ts=1675088015.554689&cid=C0195MVCEUD

tekton/pipeline-templates/params/common.yml.j2
alequint: Here is what happen @durera :

When upgrade-test pipeline runs other pipelines, it will be running, in the end, the run-pipeline role  (this one [here](https://github.ibm.com/maximoappsuite/mas-tekton/tree/master/ansible/ibm/mas_tekton/roles/run_pipeline))  inside the new mas-tekton container.

That role then, needs to donwload the pipelinerun from artifactory (that is why the artifactory key). Also, the task running inside that pipeline runs mas-tekton container, which lives in artifactory (that is why the registry). Considering we use the param list to automate also the different taskdefs params inside the upgrade-test pipeline and even the task's params and stepTemplate envs (and so pass that to run-pipeline) task, I thought would make sense to have those settings in all the pipelines (just to simplify automation, as these infos will not be used by the pipeline itself, but by the runpipeline task that will eventually run them). I am treating the pipeline and the correspondent run-pipeline task basically as the same thing parameterwise, as 95% of params are the same)

What I could do here is to rename those attributes to remove all these `fvt_` prefixes for the items in common to avoid confusion (and add some comments about why they are there). What do you think?

tekton/pipeline-templates/params/fvt.yml.j2
alequint: See comment above

tekton/pipeline-templates/params/common.yml.j2
alequint: Same reason of the "fvt" parameters described below... `cli_version` is used inside run-pipeline task to download the right version of pipelineruns from artifactory.

tekton/tasks/cert-manager.yml
alequint: Done

.github/workflows/build-cli.yml
alequint: Done. All extensions revisited following our chat in slack:
- Keep `yml.j2` for all `j2` files
- Everything else `yaml` except by:
- generate-tekton playbook, kept as `yml` (as it is ansible)

Makefile
alequint: Done. See comment above

tekton/pipeline-templates/params/common.yml.j2
durera: Makes sense ... Should move/set that param specifically for that task now, rather than have it in the global/shared param definition.

tekton/pipeline-templates/params/common.yml.j2
durera: Having spent more time working on this PR today, I think the confusion comes from the naming of this ... it should really be another pair of `devops_xxx` variables, their purpose is all about defining the records in the database that are created/updated to record the results of the tekton actions.

durera (approved):

PR #131, opened 2023-01-02
[minor] Add new Mobile test suite to FVT pipeline

jonatash:
This update also integrates `generate-pipelines` playbook into the build, so that only the template files need to be checked into the repository, avoiding the need for contributors to update the template, and generate the pipeline defintions before each commit.


jonatash:
Screenshot 2023-01-09 at 09 44 05

durera (dismissed):

durera (approved):

PR #162, opened 2023-02-06
[patch] Creating a new productID for ibm-mas UI tests

joaopauloksn:
ibm-mas tests results are being overwritten by UI tests in mongodb.

We need to give it a different product_id to differentiate it in the database.

For example coreidp-saml test Suite exists on both fvt-ibm-mas and fvt-ibm-mas-ui. So the conflict occurs.


joaopauloksn:
no need anymore

PR #163, opened 2023-02-06
[patch] Fix UI test suite conflict error

joaopauloksn:
This change will prevent the current coreidp-saml test results in mongo to be replaced.
Test Suite names in fvt-ibm-mas must be unique as we are using same product_id


alequint (approved):

alequint:
looks good to me

durera (approved):

PR #171, opened 2023-02-10
[minor] Add config tool for MAS OIDC client mgmt

hquntang:
Per Joao comments, we should allow user to perform this registration, and he recommend us to publish guideline here. This PR is to provide a guideline to allow customer to register config tool as OIDC client to specific MAS server instead of automatically register it to all MAS servers once deployed.


durera (changes requested):
Instead of providing a script in the documentation that the custer has to copy into a file this should be integrated into the CLI so that the user can run `mas oidc register`,  `mas oidc unregister` and `mas oidc update` commands.  Those commands should follow the same general UX as the other commands with an interactive and non-interactive mode.

durera (dismissed):

durera (approved):

PR #178, opened 2023-02-13
[minor] Adding cli shell support for workload scaling

joaopauloksn:
As part of the workload scaling capabilities, we want to give the customer the ability to set the workload scaling configuration when installing MAS.

As this feature is only available in pre-release catalog, the configurations settings will be available only in dev mode. When not in dev mode it will just mentioned the feature is available in pre-release catalog. @durera you might want to suggest a coding change here. I kept the numbering the same so this is why you see 5. title in both modes.

https://github.ibm.com/wiotp/tracker/issues/11185


durera (changes requested):
This will need changing based on the late design change to accomodate Turbonomics ORM.

andrercm:
@joaopauloksn @durera will we still need this PR or can we close? Not much movement on this for the past 3 months.

PR #194, opened 2023-02-25
[patch] Fix tasks without devops prefix in test attributes

alequint:


alequint:
Tested fvtcore using this fix plus things from pre-master... everything worked fine and pipeline was started successfully:

image

PR #203, opened 2023-03-10
[patch] Improvements to Manage application FVT pipeline

leo-miran:
Updates to manage task definition so tasks:
- add param `mas_appws_components` to task so pipeline tasks can verify if they should be executed with current pipeline definition
- Reorder of manage fvt-apps tasks
- Adding new user creation tasks for IS


leo-miran:
Successful run of this branch. The failed task you see in the image is related to a wrong parameter that was fed into the pipeline.
image

leo-miran:
Here's a second run where I fixed the parameter causing the issue in the previous comment
image
and a close up image
image

image/cli/install-ansible/ibm-mas_devops.tar.gz
durera, line 1: I can't merge this into master with a custom build of the ansible collection inside it ... is there a PR in the ansible collection that needs to be merged before this can be?

durera (changes requested):

image/cli/install-ansible/ibm-mas_devops.tar.gz
leo-miran: No prs for the ansible colletion that I need. I only picked this because I merged the rel/master into this branch in order to get latest updates for for artifactory. This custom build came from those merges.
Should I just delete the tar file?

durera (approved):

PR #206, opened 2023-03-16
[patch] Change app config flow to be config-based

IanBoden:
The install flow for apps was a bit of a mess with lots of nested if statements. Have changed it so that it's all done based on a config, this means it's easy to add new apps or change an app so that it does or doesn't support things like airgap or SNO and to build a prereq graph correctly. Also means a list of apps can be used in future work such as detecting installed apps during uninstall.


durera:
Can you re-raise this from a branch in this repo, so we get the benefit of the build system to help verify the change?  No need for people in our team to use forks :)

PR #213, opened 2023-03-21
[minor] Support for DB2 ldap & CIS Security enhancement

andrercm:
This PR extends support from ansible-devops to allow:

- CIS enhanced security settings to be passed as parameters to the tekton pipeline.
- DB2 resource settings to be customizable and passed as parameters to the tekton pipeline.


whitfiea (approved):
LGTM. Glad you removed that db2 version default as we already set it in the role here https://github.com/ibm-mas/ansible-devops/blob/master/ibm/mas_devops/roles/db2/defaults/main.yml#L8 and don't want two defaults!

PR #218, opened 2023-03-24
[minor] Support Manage advanced database settings

sbudhira:
Added customized database configuration for Manage database:
- Schema
- Tablespace
- Indexspace

Sample CLI
-----------
Application Selection
```
Install Manage [y/N] y
Custom Subscription Channel > 8.6.x-dev
+ Create demo data [Y/n]
+ Configure JMS [y/N]
+ Customize database settings [y/N] y
Schema > maximo
Tablespace > maximo
Indexspace > maximo
```

- Added the logic for not to install/configure DB2 if BYO JDBC config is configured.


image/cli/mascli/functions/pipeline_config_db2
```diff
@@ -193,8 +193,8 @@ function config_pipeline_storage_db2() {
                 export LOCAL_MAS_CONFIG_DIR_ALREADY_CHOSEN=yes
               fi
               jdbc_cfg_file=$LOCAL_MAS_CONFIG_DIR/jdbc-$MAS_INSTANCE_ID-wsapp.yaml
-              python3 $DIR/functions/gencfg jdbc $MAS_INSTANCE_ID $MAS_WORKSPACE_ID $MAS_APP_ID workspace-application $jdbc_cfg_file --workspace-id $MAS_WORKSPACE_ID --application-id manage
-              DB2_ACTION_MANAGE=byo
+              python3 $DIR/functions/gencfg jdbc $MAS_INSTANCE_ID $MAS_WORKSPACE_ID $MAS_APP_ID workspace-application $jdbc_cfg_file --workspace-id $MAS_WORKSPACE_ID --application-id manage           
+              DB2_ACTION_SYSTEM=byo
```
durera: Why change this?  We're in the code path that is setting up the Manage Db2, so we need to set `DB2_ACTION_MANAGE` as it currently does; unless I'm missing something here!?

tekton/src/pipelines/taskdefs/apps/db2-setup-manage.yml.j2
durera, line 1: Not sure why the change ... the logic from the #217 seems valid

Run the manage db2 hack stage if:
- installing manage
- and installing a manage specific database

You have changed this to:

Run the manage db2 hack stage if:
- installing manage
- and installing a system database
- and not installing or bringing your own manage database

I don't think this new logic works in all scenarios / I don't see what was wrong with the logic currently implemented.

e.g. if I choose not to install IoT and to install Manage, and I want to use Db2u for the Manage db2.  The logic already implemented works for this, the logic you have implemented will result in this stage not running because we are not installing a system database.

durera (changes requested):

image/cli/mascli/functions/pipeline_config_db2
sbudhira: fixed

durera (approved):

PR #219, opened 2023-03-28
[patch] Fix the condition for db2 setup for Manage app

sbudhira:
Fixed the condition for db2 setup for Manage app.

Tested scenarios 
1)  DB2 install/setup & Manage in cluster
2)  BYO JDBC cfg (external DB) with Manage


tekton/src/pipelines/taskdefs/apps/db2-setup-manage.yml.j2
```diff
@@ -11,12 +11,12 @@
     - input: "$(params.mas_app_channel_manage)"
       operator: notin
       values: [""]
-    - input: "$(params.db2_action_system)"
+    - input: "$(params.db2_action_manage)"
       operator: in
       values: ["install"]
     - input: "$(params.db2_action_manage)"
```
durera: 3rd condition here is redundant ... condition 2 will only be true if db2_action_manage == "install" ... so there's no need to also add a condition where db2_action_manage is also not "byo"

```
  when:
    - input: "$(params.mas_app_channel_manage)"
      operator: notin
      values: [""]
    - input: "$(params.db2_action_manage)"
      operator: in
      values: ["install"]
```

achieves the same thing doesn't it, or am I missing something?

durera (dismissed):

tekton/src/pipelines/taskdefs/apps/db2-setup-manage.yml.j2
whitfiea: I thought it was saying that we want to do the setup for manage only if a) manage is being installed and b) it is not a BYO DB2. If it is a manage install using a BYO DB2 then you don't want to the run the setup??

sbudhira: The first two conditions were original conditions. I added third one.

whitfiea: So further clarification is that "db2_action_manage" "install" means install db2 for manage. So if that is true then BYO would always be false, so no need for the check. @sbudhira I don't think we need that third condition.

sbudhira: @whitfiea If DB2 is installed in the cluster then set up DB2. If using BYO jdbc (external DB db2, oracle or sql server ) , then don't setup DB2.

terenceq (approved):

PR #225, opened 2023-04-03
[minor] Add param mas_wipe_mongo_data

andrercm:
Add support for running the MAS data cleanup tool as part of the uninstall pipeline.


andrercm:
Tested it with fvt-personal and it works:
```
 "mas_wipe_mongo_output.stdout_lines": [
        "================================================================================",
        "Maximo Application Suite Data Cleanup Tool",
        "================================================================================",
        "",
        "",
        "If you run this tool you must uninstall MAS and all applications, and perform a",
        "fresh installation of MAS itself and all of it's applications.  Failure to do",
        "so will result in missing indexes and other configuration in datastores that is",
        "only set up at installation time.",
        "",
        "Each action will require double confirmation, to minimise risk of accidential",
        "loss of data.",
        "",
        "MAS Core Systems",
        "--------------------------------------------------------------------------------",
        "",
        "Your confirmation string is maipempuoygzmkmnunelchunhlxcicgi",
        "",
        "Silently dropping MAS core database",
        "Database drop complete",
        "Silently dropping MAS catalog database",
        "Database drop complete",
        "Silently dropping SLS licensing database",
        "Database drop complete",
        "Silently dropping MAS adoption usage database",
        "Database drop complete",
        "Silently dropping iot_mongotest_d_core database",
        "Database drop complete",
        "Silently dropping iot_mongotest_organizations database",
        "Database drop complete",
```

Before cleaning mongo along with uninstall:
image

After:
image

The only limitation with this right now, when the clean up in mongo happens, coreapi still keeps running for a while and it seems the `core` db is created at some point, but without any data. All other dbs are deleted for good. I'm discussing with João options to fix this minor problem, in any case, that's still much better than having no option at all to wipe mongo data.

whitfiea (approved):

PR #228, opened 2023-04-04
[patch] Fixes for MongoCfg generation with existing IBMCloud MongoDb

andrercm:
This PR just extend env vars and parameters for the ansible devops PR:
https://github.com/ibm-mas/ansible-devops/pull/748


tekton/src/tasks/dependencies/sls.yml.j2
```diff
@@ -27,7 +30,7 @@ spec:
       default: ""
     - name: sls_mongodb_cfg_file
       type: string
-      default: ""
+      default: "/workspace/configs/mongo-mongoce.yml"
```
durera: Not sure this is a safe default ... in our pipeline, yes it works, but this task needs to be useable in a wider context.  What's the reasoning behind this change?

andrercm: I guess we can leave it blank as long as we set it to default in the `ansible-fvt` then?

As part of this PR I extended support for us to be able to set `sls_mongodb_cfg_file` to be something else than the mongoce config in the tekton side, something like `SLS_MONGODB_CFG_FILE=/workspace/configs/ibmcloud-mongo-$IBM_MONGO_NAME-cfg.yml`, if we want IBM Cloud Mongo config instead.

Without adding the defaulted, `none` was being passed to the pipeline while regression testing using mongoce config, so it was failing for me while installing SLS as it needs a mongo config to be used.

andrercm: We already have it here:
https://github.com/ibm-mas/cli/blob/master/tekton/src/params/install.yml.j2#L10
So didn't think it was a problem to normalize it in the sls task itself.

durera: That param is specific for our pipeline, this task is generic and could be used outside our pipeline, and in someone else's pipeline they may have the config elsewhere for instance.

durera (approved):

PR #242, opened 2023-04-18
[minor] Enable HPU to be installed via cli path

andrercm:
This PR enables Health and Predict Utilities installation via mas-cli path.


andrercm:
Tested with my own branch of cli:
image

andrercm:
As per David:
```
this path was removed after discussion with PM .. we don't want to encourage install of HPU because it's going away.  Users can still install HPU through the UI, but the less people install HPU right now the better, because we are offering minimal upgrade / migration support because HPU customers are measured in the tens.
```

So I'm closing the PR.

PR #248, opened 2023-04-19
[patch] renaming product_id and test_suite on mobile yml

jonatash:
Changing product_id from ibm-mas-mobile-xxx to ibm-mobile to all mobile products and simplifying test_suite naming in mobile yml. These changes organize the way we see the results of executions on the dashboard, encompassing mobile products in a single node.


whitfiea (changes requested):
@jonatash Can you please add a description in the PR for this change please.

whitfiea (approved):

PR #254, opened 2023-04-21
[PATCH] add creating pre and post upgrade pipelines for fvt

IanBoden:
Added preupgrade and postupgrade pipelines for use in upgrade testing. The upgrade with fvt pipeline itself is not stable so the pipelines have not been added but can be used manually.

We were creating the pipelinerun templates but not the actual pipelinerruns so added steps to produce them, with some noddy defaults to allow faster development of pipelines.


whitfiea (commented):
I think this would be a minor rather than a patch. Although it looks like I would like to defer to @alequint to review.

alequint:
Upgrade testing has changed significantly last days after the refactor made by @durera... I would defer to David review this item.

durera:
I don't think we need these changes with the simplifications made to update/upgrade/rollback/uninstall FVT pipelines.  The verification of successfulupdate/upgrade/rollback/uninstall must be directly introduced into the main pipelines .. it's a very different model to "install with integrated FVT" versus "install without FVT".

PR #261, opened 2023-04-24
[patch] Enhancements for Event Streams support

andrercm:
Adds new parameters for the following:

- Event Streams support if we don't want defaulted AMQ Kafka
- Data Dictionary channel/catalog source to be customized
- Manage base and secondary languages


andrercm:
Testing is ongoing with this, so still need to remove that ansible tar.gz, when it's done I'll set this ready to review.

image/cli/mascli/templates/pipelinerun.yaml
```diff
@@ -175,6 +190,12 @@ spec:
       value: '$MAS_CATALOG_SOURCE'
     - name: mas_catalog_version
       value: '$MAS_CATALOG_VERSION'
+    - name: mas_add_channel
+      value: '$MAS_ADD_CHANNEL'
+    - name: mas_add_catalog
+      value: '$MAS_ADD_CATALOG'
+    - name: mas_catalog_version
```
whitfiea: duplicate of line 191-192 above

image/cli/mascli/templates/pipelinerun.yaml
```diff
@@ -175,6 +190,12 @@ spec:
       value: '$MAS_CATALOG_SOURCE'
     - name: mas_catalog_version
       value: '$MAS_CATALOG_VERSION'
+    - name: mas_add_channel
+      value: '$MAS_ADD_CHANNEL'
+    - name: mas_add_catalog
```
whitfiea: these mas_add_* are also added lower day rather than here

whitfiea (changes requested):

image/cli/mascli/templates/pipelinerun.yaml
andrercm: oh thanks! i'll remove that and also:
```
    - name: mas_catalog_version
      value: '$MAS_CATALOG_VERSION'
```

which is also duplicate of line 191

image/cli/mascli/templates/pipelinerun.yaml
andrercm: Removed the dups from `/templates/pipelinerun.yml`

whitfiea (approved):

PR #262, opened 2023-04-26
[minor] Support Cloud Pak for Data 4.6.4

andrercm:
- This PR adds support for SPSS and Openscale to be installed as part of CPD
- Openscale is required for Predict app
- SPSS is optional for Predict app, therefore this will only be installed if Predict is installed and if `cpd_install_spss: True`
- Add condition while installing Predict app, to prompt for confirm if SPSS wants to be installed.
  When `CPD_INSTALL_SPSS=true` then:
  
image

   When `CPD_INSTALL_SPSS=false or null` then:

image

What might be missing, and we need decide if/when we want to include:

- This needs a new curated catalog to be added, while testing I'm using `v8-cpd464-amd64` catalog source.


whitfiea (approved):

whitfiea (changes requested):
Should the ibm-mas_devops.tar.gz be removed?

whitfiea (dismissed):
Approving but can you update the PR title as it doesn't seem to be able supporting CPD 4.6.4 but about installing SPSS and Openscale.

whitfiea:
@andrercm Also you need a changelog entry https://github.com/ibm-mas/cli/blob/master/docs/changes.md. note that 4.1 was missed and needs adding.

durera (dismissed):

PR #270, opened 2023-04-28
[patch] this is to fix grafana deployment.

ctrafa:
This is to fix grafana deployment error found, where storage class and size has the same value!


ctrafa:
Already fixed by @durera

PR #287, opened 2023-05-09
[patch] Separate Manage and Manage IS FVT results in two different groups in dashboard

alequint:
This part was supposed to be included in the last minor change (4.2) but a bad merge using old product id frustrated that. Fixing here.

@leo-miran could pls review the PR as well before merging?


durera (changes requested):
Use a common prefix in the suite names, ibm-mas-manage-is is not a valid productid, it must track back to the operator name/case bundle ID, which for manage and all it’s components is ibm-mas-manage.

alequint:
Move PR to draft as tasks have problems of duplicated params in case of product_id according to the latest tests:

```
Fatal: [localhost]: FAILED! => {"attempts": 10, "changed": false, "msg": "Failed to create object: b'{\"kind\":\"Status\",\"apiVersion\":\"v1\",\"metadata\":{},\"status\":\"Failure\",\"message\":\"admission webhook \\\\\"validation.webhook.pipeline.tekton.dev\\\\\" denied the request: validation failed: params names must be unique, the same param: product_id is defined multiple times at: spec.tasks[63].params[15].name, spec.tasks[64].params[15].name, spec.tasks[65].params[15].name, spec.tasks[66].params[15].name\",\"reason\":\"BadRequest\",\"code\":400}\\n'", "reason": "Bad Request"}
```

alequint:
> Use a common prefix in the suite names, ibm-mas-manage-is is not a valid productid, it must track back to the operator name/case bundle ID, which for manage and all it’s components is ibm-mas-manage.

@durera this is how it is today. The idea here is because is too difficult to separate what is Manage core tests (for Manage core developers) from the Industry Solutions developers, which is a different thing in Manage backstage. Even though it does not relate to a valid operaror/product it, it falls into the same situation of Mobile tests, as ibm-mas-mobile is not a valid application as well, but that id is used to separate Mobile tests results from the rest.

alequint:
Covered by #289

PR #307, opened 2023-05-19
[minor] Add MAS Core backup & restore script to CLI

terenceq:
### Convenience Script for MAS Core backup/restore

A simple bash script to assist with backing up resources in the MAS Core namespace. The script only backs up resources that would be required when restoring the same identical instance (i.e. same cluster, instance, etc)


image/cli/mascli/backup-restore/mascore-backup-restore.sh
durera, line 1: Can you add the backup-restore directory to the Dockerfile and make sure the script is executable in the container too?

durera (changes requested):

image/cli/mascli/backup-restore/mascore-backup-restore.sh
terenceq: It should already be added to the image via this line in the Dockerfile:

```
COPY --chown=default:root mascli /mascli
```

I've added an update to set the contents of `backup-restore` to executable.

```
chmod +x /mascli/backup-restore/* && \
```

PR #308, opened 2023-05-19
[minor] Add support for AWS Route 53 as DNS provider for MAS installs

andrercm:
This PR enables option to configure an existing (it does not create one) AWS route53 hosted zone and configures a custom cluster issuer and corresponding webhook to generate Let's Encrypt well known public certs for MAS.

image

This PR also fixes couple of problems identified during AWS provisioning:

- `AWS_SECRET_KEY and AWS_ACCESS_KEY_ID` re-use saved features replaced to use common function `prompt_for_secret`
- Fixed `prompt_for_input` path to properly read `IPI_PULL_SECRET_FILE` variable. Without the fix there's no way to set this var as this is using a default path/filename, and it will fail if you don't have the defaulted file in the expected path: `/opt/app-root/src/masconfig/pull-secret` which likely you will only find out if you ended up failing the first time. With the fix you will get prompted to provide the file path for the aws pull secret file just like we do for SLS entitlement file, and you get an error right away if the file path is invalid.

Corresponding changes in ansible-devops: https://github.com/ibm-mas/ansible-devops/pull/828
Corresponding changes in ansible-fvt: https://github.ibm.com/maximoappsuite/ansible-fvt/pull/217


image/cli/mascli/functions/provision_aws
```diff
@@ -83,7 +83,7 @@ function set_default_vars() {
   export IPI_CONFIG_DIR="/mascli/masconfig"
   export IPI_PLATFORM=aws             # IPI_PLATFORM
   export CLUSTER_TYPE=ipi             # CLUSTER_TYPE
-  export IPI_PULL_SECRET_FILE="/opt/app-root/src/masconfig/pull-secret"                   # IPI_PULL_SECRET_FILE
+  # export IPI_PULL_SECRET_FILE="/opt/app-root/src/masconfig/pull-secret"                   # IPI_PULL_SECRET_FILE
```
whitfiea: Can we just delete this line?

whitfiea (changes requested):
Can you add the changelog entry as well?

image/cli/mascli/functions/provision_aws
andrercm: Done

whitfiea (approved):

PR #309, opened 2023-05-21
[minor] Add Manage backup and restore script to CLI

sbudhira:
Convenience Script for the Manage Application Backup/Restore
This is a simple bash script designed to assist in backing up resources within the Manage namespace. The script selectively backs up only the resources necessary for restoring an identical instance (e.g., same cluster, instance, etc.).


durera (changes requested):
Can you add the backup-restore directory to the Dockerfile and make sure the script is executable in the container too?

durera (approved):

PR #317, opened 2023-05-25
[minor] Add Support for May 26 Catalog Update

just4li:
The May 26 catalog update delivers a critical fix for Monitor 8.10.x subscription channel, the Monitor 8.10.1 release contained packaging errors that result in it not being usable in both connected and disconnected environments.


whitfiea (commented):
@just4li The PR needs a description. I am not sure we want to replace the existing catalog. We have had customers ask before why the static catalogs disappear some times, and this would be such a case. I understand that the reason is due to the Monitor 8.10.1 not being usable due to a digest issue, but the other apps are ok. I would think only if the catalog is completely useless then we would want to replace it rather than add this as just a newer catalog. A comment in the docs to say don't use catalog v8-230518-amd64 if you want to use Monitor would also help. My recommendation is to update the docs/ctalog/index.md to add this new catalog rather than replace the previous May one

mkdocs.yml
```diff
@@ -27,7 +27,7 @@ nav:
       - "configtool-oidc": commands/configtool-oidc.md
   - "Catalogs":
       - "Overview": catalogs/index.md
-      - "May 18 2023": catalogs/v8-230518-amd64.md
+      - "May 26 2023": catalogs/v8-230526-amd64.md
```
whitfiea, line 30: Needs to keep the       - "May 18 2023": catalogs/v8-230518-amd64.md I believe

durera: I was thinking just listing the "main" ones in the side nav (one per month, and only the last 4 months), the `catalogs/index.md` page has the full list with no ommissions (or should anyway :) )

whitfiea (approved):

PR #328, opened 2023-06-09
[patch] Fix for IoT installation not detecting storage class

andrercm:
The issue here is that the auto detect logic might not find any available default storage classes, and we dont have anything blocking the installation from beginning if messagesight storage classes are not defined. Therefore this adds an assertion that prevents the iot install from beginning if iot size != dev and storage classes for fpl and message sight are undefined.
So fixing the cli to use the common block storage classes for iot related storage classes to ensure it is also passed to iot install ansible roles


andrercm:
Running a test from cli:
image

IoT got installed fine:
image

PVCs correctly bound:
image

whitfiea (approved):

PR #329, opened 2023-06-10
[minor] Support mix of in-cluster & BYO databases

durera:
With this update it is possible to choose to mix bring your own and in-cluster databases during the installation.

image

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -1,7 +1,9 @@
 ## Changes
 
 
-- [`5.3`](https://github.com/ibm-mas/cli/releases/tag/5.3.0) Support affinity and tolerations for Db2 install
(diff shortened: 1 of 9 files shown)


whitfiea (commented):
@durera how does this work for Monitor and Predict? as I don't see any option for them?

durera:
> how does this work for Monitor and Predict? as I don't see any option for them?

Basically the same as before the change:

- Monitor can only be used if IoT has been selected, and always uses IoT's database - does not support having it's own (which is why IoT can only support system scope database)
- Predict without IoT I don't know off the top of my head as I don't know too much about the internals of what Predict does/doesn't do with databases .. but it's unchanged / works the same as it worked before ... Predict is gated only by the Manage install, and doesn't need it's own database with or without IoT.

whitfiea (approved):

PR #330, opened 2023-06-13
[minor] Manage test phases and additional FVTs + first Manage x Monitor IVT task

alequint:
- Additional Manage core tests
- First IVT Task, with 4 steps: 3 for Manage and 1 for Monitor (`mas-ivt-manage-monitor-dd`) testing, together, Data Dictionary, where data is generated in Manage.
- Reorganization of Manage tests in 4 different phases as some tests must run after others and others must run in isolation


tekton/test.sh
andrercm, line 1: Why is that adding an empty file?

andrercm (changes requested):
just a minor question

tekton/test.sh
alequint: I have no idea why it is showing as empty tbh... maybe because it is a sh? weird... and I did not even modify this file, not sure why it is here. You can see the content by :

image

I will review why it is in he PR anyway... btw, hold on PR because I forgot changelogs as always :)

alequint: @andrercm added a new line in the end of test.sh, just to have something here, as the file is not empty at all :) 

I have also updated change logs as this is a mino version

PR #332, opened 2023-06-13
[patch] sps manage

ypkim-github:
- Triggering the dynamic scan for manage on SPS
- All same with https://github.com/ibm-mas/cli/tree/master/tekton/src/pipelines/taskdefs/ivt-core except for calling `ibmcloud dev tekton-trigger`

### DIFF
--- a/tekton/src/pipelines/taskdefs/sps-core /common /params.yml.j2
+++ b/tekton/src/pipelines/taskdefs/sps-core /common /params.yml.j2
@@ -1,0 +1,10 @@
+- name: mas_instance_id
+  value: $(params.mas_instance_id)
+- name: mas_workspace_id
+  value: $(params.mas_workspace_id)
+- name: fvt_image_registry
+  value: $(params.fvt_image_registry)
+- name: fvt_image_name
+  value: ivt-core
+- name: fvt_image_digest
+  value: $(params.ivt_digest_core)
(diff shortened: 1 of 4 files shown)


whitfiea (dismissed):
@ypkim-github this needs a description of that this PR does and how it is tested etc. You also need to make sure it doesn't use personal image builds like `fvt-ibm-mas-manage:2.6.1-pre.vncupdates`

durera (commented):
In addition to Andrew's comments, looking at the changes I don't think this will work in fvt-airgap.

- use of the ubuntu image will require that is mirrored too and a ISCP set up for it, would be better off using the CLI image seeing as it already have the tools you are installing onto the ubuntu image anyway
- use of hardcoded artifactory for the registry won't work because artifactory is blocked

ypkim-github:
@durera @whitfiea

1. Any idea with this build-cli error?
 ```
Error parsing reference: "quay.io/ibmmas/cli:" is not a valid repository/tag: invalid reference format
```
2. How do we get the public endpoint of the cluster that is just provisioned? e.g., $(params.fvt_extenernal_host)
3. How do we make this task run only on the FVT clusters?

whitfiea:
@ypkim-github the error `Error parsing reference: "quay.io/ibmmas/cli:" is not a valid repository/tag: invalid reference format` is becuase you have no commit messages with a [patch] (just the PR title has it). So no version can be determined.

I don't know about the other questions. You would need to ask @alequint

PR #336, opened 2023-06-16
[minor] Manage and Monitor FVT adjusts

alequint:
Several changes from Squad CI Sprint work:

- **Rearrange MIF Tests** to accomodate Data Dictionary data where Monitor creates data ("monitor first scenario")
we need to run mif-setup keyword before every other MIF test, that is one change, the second change is that the mif-jms-setup is now being covered by the mif-setup, so that keyword is not necessary anymore, the 3th change is the mif-dd-setup keyword is not necessary anymore as the setup script will be called from within the dd test script (by @creyesibm )
- **Fix Monitoring task to use pytest container** We are not getting result on fvt dashboard for fvt-manage-pytest as it does not consider right digest from fvt-manage-pytest. These changes will point to correct digest for pytest. Refer to https://github.ibm.com/wiotp/tracker/issues/12802#issue-34609524 (by @unnati-solanki-git )
- **Direct Print keyword**: added keyword changes for the direct print report as well. Refer to https://github.ibm.com/wiotp/tracker/issues/11914#issue-33174367 (by @unnati-solanki-git )
- **New Monitor FVT task** to run when Manage is installed (fvt-monitor-manage) (by @alequint )
- **Pytest container fix**: https://github.com/ibm-mas/cli/pull/350 (by @unnati-solanki-git )
- **Move manage x monitor IVT task** to Monitor section in FVT Dashboard (by @alequint )

----

Most recent run in https://dashboard.masdev.wiotp.sl.hursley.ibm.com/tests/pfvtsquadci2. Note that failures are being worked as we speak and monitored in a daily scrum. This PR is not introducing new failures, as most of the test failing here are already in there, but it contains several adjustments (in DD area and in manage pytest task) that we woudl like to merge before we accumulate more. pfvtsquadci12 had a relatively good run after a long a time struggling

image

----

July 5th run

image

Some failures still there, others fixed, new ones appearing... but from a pipeline perspective it is running fine till the end, we just need to continue investigation of failures and identify if problem is in the test or in the application now 

---- 

Slack Thread: https://ibm-watson-iot.slack.com/archives/C031Q7W21FZ/p1688027685661419

PR #341, opened 2023-06-20
[patch]pytest fixes and direct print keyword

unnati-solanki-git:
We are not getting result on fvt dashboard for fvt-manage-pytest as it does not consider right digest from fvt-manage-pytest. These changes will point to correct digest for pytest. 

https://github.ibm.com/wiotp/tracker/issues/12802#issue-34609524

Also, added keyword changes for the direct print report as well

https://github.ibm.com/wiotp/tracker/issues/11914#issue-33174367


alequint:
@unnati-solanki-git I did incorporate your changes in this PR: https://github.com/ibm-mas/cli/pull/336 - simpler to handle our pfvtsquadci runs

PR #350, opened 2023-06-27
[patch]pytest changes with new task

unnati-solanki-git:
We are not getting result on fvt dashboard for fvt-manage-pytest as it does not consider right digest from fvt-manage-pytest. These changes will point to correct digest for pytest.

https://github.ibm.com/wiotp/tracker/issues/12802


tekton/src/pipelines/taskdefs/fvt-apps/manage.yml.j2
```diff
@@ -20,7 +20,7 @@
 
 # Manage FVT Monitoring
 {# - name: fvt-manage-monitoring
```
alequint: @unnati-solanki-git task is commented out, I will bring it back and merge... everything else looks good to me :+1:

alequint (approved):

PR #358, opened 2023-07-10
[patch] Manage FVT Pipeline updates

unnati-solanki-git:
The following changes are covered: 
- Added SMTP keyword #12924  (by @unnati-solanki-git )
- Added Manage User Sync - added by Xinyu (who tested this version of pipeline end to end) (by @alequint )
- Moved metrics-api to phase 2 as it depends on setup as per @unnati-solanki-git review (by @alequint )
- Added wo attach doc test in phase 3 (by @alequint ): `wo-configure-attach-docs`

Pipeline:

image


tekton/src/pipelines/taskdefs/fvt-apps/manage.yml.j2
```diff
@@ -239,3 +239,14 @@
   # Update runAfter list with new PHASE 4 tests
   runAfter:
     - fvt-manage-startcenter-config
```
alequint: @unnati-solanki-git I have noticed Direct Print is being tested in paralel with Work flow designer (which was supposed to run isolated, like start center) - and I have noriced, btw, my comment in line 211 is wrong, indicating start center is the last test... the last one should be workflow designer (all the Indudtry solution tests are configured to run after workflow).

We need to add direct print and smtp in a different place... do they have any restriction or can run in parallell with any test? Because if so, we can add them in different test phases (I will improve that by separating the different phases in different files later anyway).

Also, be sure to update comments on top of the different test phases, with the new tests we are adding.

alequint (changes requested):

tekton/src/pipelines/taskdefs/fvt-apps/manage.yml.j2
unnati-solanki-git: Hi @alequint direct-print and smtp both have dependency on setup only. I agree we should add them in different test phases.

alequint: @unnati-solanki-git I did reorganize the pipeline and now let smtp as the latest task instead of workflow designer and move other tests to different phases (as I've noticed not all the tests depend on setup). It is already in this PR, but here is the commit FYI: https://github.com/ibm-mas/cli/commit/0ba73a547c3a50d5983fa0236eaa9576d7f68a42

tekton/src/pipelines/taskdefs/fvt-apps/manage.yml.j2
```diff
@@ -45,8 +54,36 @@
     - name: fvt_test_suite
       value: metrics-api
   runAfter:
-    - fvt-manage-setup
```
unnati-solanki-git, line 48: @alequint fvt-manage-metrics-api needs to run after fvt-manage-setup as it requires the super user to be set up before running the script. 

Checkout result of pfvtsquadci12 here: https://dashboard.masdev.wiotp.sl.hursley.ibm.com/tests/pfvtsquadci12/testsuite/ibm-mas-manage/metrics-api

image

alequint: Oh i see.... good catch @unnati-solanki-git, my criteria was the presence of SetupProperties.data as dataInput and I did not see it for metrics-api. But it seems criteria is not that good, and need to review other "Phase 1" tests, if they require the same.

alequint: From PHASE 1: all the following succeed or we know why failed:
- fvt-manage-monitoring
- fvt-manage-security-audit-logging
- fvt-manage-security-inactive-auth

...so metrics-api is the only one to be moved to other phases :+1:

alequint (changes requested):
move metrics-api to run after phase 1 test as it requires maxadmin credentials

PR #376, opened 2023-07-21
[minor] Adding v3 user data model tests

durera:
### Description
Adding a new set of tests for the new usermgmt-v3 apis.

This steps will run similar tests as usermgmt but using new MAS user data model.

These tests will be skipped for 8.9.x and 8.10.x environments.


tekton/src/pipelines/taskdefs/fvt-core/phase2.yml.j2
```diff
@@ -1,12 +1,12 @@
 # Only start phase 2 tests after all suites in phase 1 are complete
 
-# Core FVT 2.1 coreapi-mobileapi ~5m
-- name: fvt-coreapi-mobileapi
+# Core FVT 2.0 coreapi-usermgmt-v3 ~30m
```
alequint, line 3: So, now phase 2 will be like:

After run `phase1`, the following tests will run in parallel:
- fvt-coreapi-usermgmt-v3
- fvt-coreapi-dependencies
- fvt-coreapi-messages

And after `fvt-coreapi-usermgmt-v3`, the following wil run parallel:
- fvt-coreapi-mobileapi
- fvt-coreapi-other
- fvt-coreapi-utilities
- fvt-coreapi-uiresources

Is that is the idea, looks good 👍

alequint (approved):
Refer to: https://github.com/ibm-mas/cli/pull/376#discussion_r1283714477

PR #377, opened 2023-07-21
[minor] Support July Catalog Update

sanjayprab:
For July catalog - https://github.ibm.com/maximoappsuite/ibm-maximo-operator-catalog/pull/260

This update removes earlier catalogs from the selection prompt, due to the problems with the Postgres Operator used by IBM User Data Services we can not realistically support customers on catalogs prior to the July 21st update.  Eventually they will pick up the breaking update in their Red Hat Certified Operator Catalog, which will break UDS.

image


durera (changes requested):
As discussed in Slack

durera (approved):

PR #394, opened 2023-08-03
[patch] removing custom ansible devops tar

leo-miran:
Removing custom ansible devops tar in order to pick up latest version.


durera:
Fixed in another PR just merged ...

PR #417, opened 2023-08-14
[minor] Support setting "default CA trust" during install

stonepd:
Pass the MAS_TRUST_DEFAULT_CAS environment variable to the install FVT pipelines.

This has been tested in personal FVT cluster pfvtpds


docs/commands/install.md
```diff
@@ -191,14 +192,20 @@ If you choose the latter then you will be prompted to select a local directory w
     If you have already generated the configuration file (manually, or using the install previously) the CLI will detect this and prompt whether you wish to re-use the existing configuration, or generate a new one.
 
 
-### Step 10. Additional Configurations
+### Step 10. Configure Turbonomic
```
leo-davison, line 195: @stonepd This section seems to have the same text as the next section?

Also, if you view the file (outside of the diff) the numbering on these headers isn't quite right

leo-davison (dismissed):

durera (approved):

docs/commands/install.md
durera: I'll fix this in my branch ... looks like I started to update the doc for Turbonomic, but didn't finish it

durera (approved):

PR #429, opened 2023-08-21
[minor] Add ROSA deprovision pipeline

whitfiea:
Adds the following new functions:
- gitiops-deprovison-mongo
- gitops-deprovision-cluster
- gitops-deprovsion-rosa

Also then includes that tekton tasks for these and a new pipeline called `deprovsion-cluster` that will call these tasks.

The pipeline will delete the docdb instance, remove the details from secrets-manager. It will then remove the cluster resources like ibm-common-services and then finally deletes the rosa cluster and infrastructure around it.

As a sidechange this also updates the mas-provision tasks so that we don't need to provide the vpc_id but instead computes it based on the name of the cluster.

This is all part of the work for internal issue https://github.ibm.com/wiotp/tracker/issues/13189

It has been tested and confirmed to work, there was an issue with the mongodb role not waiting long enough after sending the delete request but this has been updated here https://github.com/ibm-mas/ansible-devops/commit/7ae5668f0d54f861b6e3d586afc2be713c3fde62

### DIFF
--- a/image/cli/mascli/functions/gitops_deprovision_cluster
+++ b/image/cli/mascli/functions/gitops_deprovision_cluster
@@ -1,0 +1,232 @@
+#!/usr/bin/env bash
+
+function gitops_deprovision_cluster_help() {
+  [[ -n "$1" ]] && echo_warning "$1"
+  reset_colors
+  cat << EOM
+Usage:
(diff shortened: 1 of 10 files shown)


padmankosalaram (approved):
Please merge the changes

Issue #431, opened 2023-08-21
missing dependencies used by MAS Core and Manage backup and restore


terenceq:
## MAS Core and Manage backup and restore dependencies

The MAS Core and Manage backup procedures document to use the following tools:

- Mongo Shell
- mongodump
- mongorestore
- yq

The CLI container should control the install and version of these tools so compatibility can be maintained.

durera:
Closing issue.  This was delivered on Aug 21st in #432

PR #432, opened 2023-08-21
[minor] Add dependencies used by MAS Core and Manage backup and restore

terenceq:
## Ensure MAS Core and Manage backup and restore dependencies are installed

https://github.com/ibm-mas/cli/issues/431

For compatibility the following dependencies/tools will be installed and configured:

- Mongo Shell
- mongodump
- mongorestore
- yq


image/cli/install/install-yq.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+# Install yq CLI
+set -e
+
+curl -L  "https://github.com/mikefarah/yq/releases/download/v4.35.1/yq_linux_amd64"  > yq
```
whitfiea: The previous yq was this one https://pypi.org/project/yq/ (https://github.com/kislyuk/yq) which is different to https://github.com/mikefarah/yq is there a reason for the change? I believe we are using the original yq in places

whitfiea (changes requested):

image/cli/install/install-yq.sh
terenceq: I cannot find any references to this and that is why I removed it completely:

```bash
# Install yq
# ----------
function install_yq() {
  python -m pip install yq || exit 1
}
```

I feel the version of `yq` from `mikefarah` is better than the Python one. Also that is the one we used to develop the backup and restore scripts.

terenceq: I suppose I could also leave that `install_yq` in the build script. Not sure if it would ever be used though. I removed it to avoid any confusion.

whitfiea: @terenceq it is being used in a few places in the gitops branch like here https://github.com/ibm-mas/cli/blob/gitops/image/cli/mascli/functions/gitops_init_suite#L323 and also the prereq installed jq that comes with it. We choose that as that is what is installed.

terenceq: @whitfiea ok - I put the `mikefarah` version under `/opt/mikefarah/bin` and I'll introduce some logic in the backup/restore scripts that will allow setting the path to the yq binary in a later PR.

whitfiea (approved):

PR #434, opened 2023-08-22
[patch] adding fvt-mobile-standalone pipeline

leo-miran:
This new pipeline is focused on mobile only fvt tests. Since `mas-install-with-fvt` pipeline has mobile tests at the very end, it takes quite a while to verify mobile results and often get blocked due to unrelated problems elsewhere in the pipeline.


leo-miran:
Here's evidence of a successful run of this pipeline
[mas-pfvtmob2-pipelines.txt](https://github.com/ibm-mas/cli/files/12408519/mas-pfvtmob2-pipelines.txt)

leo-miran:
Need to fix a problem I found. Will open a new PR

PR #436, opened 2023-08-23
[patch] Delete config resource via oc cli

padmankosalaram:
- [ ] delete the configs via oc cli as argocd is not deleting them due to owner reference


whitfiea (approved):
LGTM and glad you added the comment to say this is a workaround so we don't forget it.

PR #437, opened 2023-08-23
[patch] Add Maximo IT license & support for --accept-license flag

chriscochran:
As a part of Maximo IT add-on for Manage, we must add the terms and agreements for the separately licensed product. For the MAS 8.11 release we don't require to save the customer's response to a MAS or Manage CR. 

This PR should not be merged until the product announcement date.

**Notice:** The snipped url for Maximo IT v8.1 license is temporarily associated with the target url: https://www.ibm.com/products/maximo for test purpose only. And when we achieve MAS v8.11 announcement date and Maximo IT's license public url is generated, I will replace the target url with the final public url. You don't need any change from your side, and the final public url of Maximo IT license will be re-directed and take effective.


image/cli/mascli/functions/install
```diff
@@ -174,6 +174,11 @@ function install_noninteractive() {
     export MAS_APP_CHANNEL_MONITOR=$1 && shift
     ;;
     --manage-channel)
+    echo
+    echo_h2 "Maximo IT Add-On"
+    echo -e "${COLOR_YELLOW} For information about your license, see https://ibm.biz/MAXIT81-License You are not required to accept the license terms unless you are installing Maximo IT add-on for Manage."
+    prompt_for_confirm_default_yes "Do you accept the license terms?" LICENSE_RESPONSE
```
whitfiea: This is part of the noninteractive install so I am not sure you human would be expected to be prompted?

image/cli/mascli/functions/pipeline_config_applications
```diff
@@ -152,6 +152,11 @@ function config_pipeline_applications() {
       prompt_for_input "Tablespace" MAS_APP_SETTINGS_TABLESPACE
       prompt_for_input "Indexspace" MAS_APP_SETTINGS_INDEXSPACE
     fi
+    echo
+    echo_h2 "Maximo IT Add-On"
+    echo -e "${COLOR_YELLOW} For information about your license, see https://ibm.biz/MAXIT81-License You are not required to accept the license terms unless you are installing Maximo IT add-on for Manage."
+    prompt_for_confirm_default_yes "Do you accept the license terms?" LICENSE_RESPONSE
```
whitfiea: Is there not an option in the cli to install the IT addon? It would be nicer to have an "install Maximo IT addon" section and then display this prompt. We can then fail if no is selected.

whitfiea (changes requested):

image/cli/mascli/functions/pipeline_config_applications
chriscochran: Hi @whitfiea from what I understand there will be an option to install Maximo IT add-on in the future. Right now I will make changes so the user has to accept both licenses.

image/cli/mascli/functions/install
chriscochran: Thanks Andrew, I removed this part.

chriscochran:
Update: Got confirmation that we don't need to wait until the announcement date to close this PR.

chriscochran:
Screenshot 2023-08-30 at 1 41 12 PM

whitfiea (dismissed):
The prompt seems good now. Approving

whitfiea:
Needs to merge when the link to the license will actually point to the license. We can't tell user to approve the license and then not direct them too it.

image/cli/mascli/functions/pipeline_config_applications
```diff
@@ -137,6 +137,10 @@ function config_pipeline_applications() {
 
   # Manage
   if prompt_for_confirm "Install Manage"; then
+    echo "Maximo IT License Terms"
```
durera: Please move this logic into the step where we configure Manage components.  We can use the components the use selected to determine if Maximo IT is being installed (it's just a component in Manage right?) This will allow us to show this only when it's meaningful to the user (they have added the Maximo IT component to the install).

durera (dismissed):

chriscochran:
Added `--accept-license` flag to non-interactive installation and updated the docs:

Screenshot 2023-09-14 at 3 32 54 PM


If user has added Maximo IT to the Manage components list in interactive mode, they will be prompted to accept the license:

Screenshot 2023-09-13 at 2 53 02 PM


If the user has added Maximo IT to the Manage components list and doesn't accept the license, the program will exit: 

Screenshot 2023-09-13 at 2 52 46 PM


If the user hasn't added Maximo IT to the Manage components list the license accept prompt doesn't appear: 

Screenshot 2023-09-13 at 2 52 26 PM

image/cli/mascli/functions/install
```diff
@@ -386,6 +388,9 @@ function install_noninteractive() {
     --uds-lastname)
     export UDS_CONTACT_LASTNAME=$1 && shift
     ;;
+    --accept-license)
+    export LICENSE_ACCEPTED=true && shift
```
whitfiea: If this is just a boolean flag i.e. `--accept-license` with no value then we don't need the shift at the end. This should probably be under the "Other commands" section below https://github.com/ibm-mas/cli/pull/437/files#diff-bd207331dfe68009e28fb66d98bb4719786c180b4716ed61ab76d82105179f7dR471

whitfiea (changes requested):

durera:
Placed in draft until the final URL for the license is available

chriscochran:
Screenshot 2023-09-26 at 9 33 40 AM
 URL is ready. Good to merge.

whitfiea (approved):

PR #440, opened 2023-08-25
 [patch]CIS Compliance for SaaS #12281 

Yolandabjyf:
- create a new function gitops_init_cis_compliance for CIS Compliance 
- add argocd application template for this

story link: https://github.ibm.com/wiotp/tracker/issues/12281


image/cli/mascli/functions/gitops_init_cis_compliance
```diff
@@ -0,0 +1,206 @@
...
+      -c|--cluster-id)
+        export CLUSTER_ID=$1 && shift
+        ;;
+
+      # Secrets Manager
+      --secrets-path)
+        export SECRETS_PATH=$1 && shift
+        ;;
+      --secrets-key-seperator)
+        export SECRETS_KEY_SEPERATOR=$1 && shift
```
whitfiea: I don't think this function needs the secrets related options does it?

image/cli/mascli/templates/gitops/cis-compliance.yaml.j2
```diff
@@ -0,0 +1,27 @@
...
+  source:
+    repoURL: https://github.com/ibm-mas/gitops
+    path: applications/cis-compliance
+    targetRevision: {{ GITOPS_VERSION }}
+  syncPolicy:
+    automated:
+      prune: true
+      selfHeal: true
+    syncOptions:
+      - CreateNamespace=true
```
whitfiea: We should set this to false as we don't want it to create namespaces

whitfiea (changes requested):

image/cli/mascli/functions/gitops_init_cis_compliance
Yolandabjyf: Yes, you are right. I have removed the secret manager part.

image/cli/mascli/templates/gitops/cis-compliance.yaml.j2
Yolandabjyf: I have change it to false. In what situation we don't want it to create namespaces?

whitfiea: We found a few issues like argocd won't delete the namespace that it created when the CreateNamespace=true is set. Instead we add the namespace resource into the application like this https://github.com/ibm-mas/gitops/blob/poc/applications/ibm-mas-suite/templates/01-ibm-mas_namespace.yaml so it is just another resource.

tekton/src/tasks/gitops/gitops-init-cis-compliance.yaml
```diff
@@ -0,0 +1,74 @@
...
+      - |-
+        git config --global user.name "yourUsername"
+        git config --global user.email "you@example.com"
+        git config --global user.password $GITHUB_PAT
+
+        ROSA_CONFIG=/workspace/configs/tmp-rosa-init/rosa-$(params.cluster_name)-details.yaml
+        export OCP_SERVER=$(cat $ROSA_CONFIG | yq -r '.data.api_url')
+        export OCP_USERNAME=$(cat $ROSA_CONFIG | yq -r '.data.username')
+        export ROSA_CLUSTER_ADMIN_PASSWORD=$(cat $ROSA_CONFIG | yq -r '.data.admin_password')
+        oc login -u $OCP_USERNAME -p $ROSA_CLUSTER_ADMIN_PASSWORD $OCP_SERVER --insecure-skip-tls-verify
```
whitfiea, line 50: Can we add the following after this line so we exit if we can't log in. 
```
        OC_RC=$?
        if [ $OC_RC -ne 0 ]; then
          echo "oc login failed"
          exit 1
        fi
```

image/cli/mascli/templates/gitops/cis-compliance.yaml.j2
```diff
@@ -0,0 +1,27 @@
...
+    repoURL: https://github.com/ibm-mas/gitops
+    path: applications/cis-compliance
+    targetRevision: {{ GITOPS_VERSION }}
+  syncPolicy:
+    automated:
+      prune: true
+      selfHeal: true
+    syncOptions:
+      - CreateNamespace=false
+    managedNamespaceMetadata:
```
whitfiea: This `managedNamespaceMetadata` section we need to remove as it was causing issues. That change has come in recently after you did these changes. Just delete lines 25 to 27

image/cli/mascli/templates/gitops/cis-compliance.yaml.j2
```diff
@@ -0,0 +1,27 @@
+---
+# IBM Cloud Pak Foundational Services
```
whitfiea: Comment needs fixing

image/cli/mascli/functions/gitops_init_cis_compliance
```diff
@@ -0,0 +1,193 @@
...
+        ;;
+
+      # Other Commands
+      -h|--help)
+        gitops_init_cis_compliance_help
+        ;;
+      *)
+        # unknown option
+        echo -e "${COLOR_RED}Usage Error: Unsupported option \"${key}\"${COLOR_RESET}\n"
+        gitops_init_cis_compliance_help
```
whitfiea: This line needs to add the message so it becomes 
```
gitops_init_cis_compliance_help "Usage Error: Unsupported option \"${key}\" "
```
If we don't do that then the function returns with a rc of 0 and so the task completes ok when it should fail.

whitfiea (changes requested):

whitfiea (approved):

PR #448, opened 2023-08-30
[patch] adding support to pass docdb instance count as param

padmankosalaram:


whitfiea (commented):
These changes look good but can we also include the instance number on the pipelineruns that are used to run fvt. This would mean adding it here https://github.ibm.com/maximoappsuite/ansible-fvt/blob/gitops/ansible/ibm/mas_fvt/roles/start_pipeline/templates/pipelineruns/gitops-provision-cluster-after-deprovision.yml.j2#L66 and picking up the variable here https://github.ibm.com/maximoappsuite/ansible-fvt/blob/gitops/ansible/ibm/mas_fvt/roles/start_pipeline/vars/mas810saas.yml#L44. Finally setting the variable here https://github.ibm.com/maximoappsuite/fvt-dev/blob/fvtsaas/.travis.yml#L66 so it is 1 rather than the default 3?

whitfiea (approved):

PR #450, opened 2023-08-31
[patch] Fix quota type setting for fyre provision

racree:
Fix bug in provision_fyre where FYRE_QUOTA_TYPE is always set to product_group when interactive.

### DIFF
--- a/image/cli/mascli/functions/provision_fyre
+++ b/image/cli/mascli/functions/provision_fyre
@@ -148,12 +148,12 @@ function provision_fyre_interactive() {
 
   prompt_for_confirm "Use Product Group Quota?" USE_PRODUCT_GROUP_QUOTA
   if [[ "$USE_PRODUCT_GROUP_QUOTA" == "true" ]]; then
-    FYRE_QUOTA_TYPE=product_group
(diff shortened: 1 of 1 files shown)


racree:
When I reply "N" to "Use Product Group Quota?" it results in quota type of "quick_burn"
Screenshot 2023-08-31 at 4 23 42 PM
Screenshot 2023-08-31 at 4 25 45 PM

And since there is no more capacity in quick burn it correctly fails with
 "quick_burn memory quota exceeded for platform x for site svl for product group id 225"]

When I respond with "Y" to "Use Product Group Quota?" it results in quota type of "product_group"

Screenshot 2023-08-31 at 4 29 36 PM

And since my fyre user does not have OCP+ quota, it correctly fails with
{"errors": ["individual product_group allowed cpu exceeded for user_id 13891 on platform x in product_group_id 225 for site svl", "individual product_group allowed memory exceeded for user_id 13891 on platform x in product_group_id 225 for site svl", "individual product_group allowed disk exceeded for user_id 13891 on platform x in product_group_id 225 for site svl"]}

whitfiea (approved):

PR #458, opened 2023-09-06
Deprovision EFS and MSK Instance changes

padmankosalaram:
- [x] Added new functions to deprovision EFS, MSK Kafka and delete kafka config
- [ ] Added new tasks to provision / deprovision EFS, MSK kafka and kafka config
- [x] Adding above tasks to pipeline deprovision-mas-instance.yml.j2 and provision-mas-instance.yml.j2


image/cli/mascli/functions/gitops_deprovision_kafka
```diff
@@ -0,0 +1,293 @@
...
+
+
+    
+    #kafka aws provider role specific env variables
+    export MAS_CONFIG_DIR=$TEMP_DIR
+    export KAFKA_ACTION="uninstall"
+    
+    export MAS_INSTANCE_ID=${MAS_INSTANCE_ID:-CLUSTER_ID}
+    export KAFKA_NAMESPACE="msk-${CLUSTER_ID}"  
+    export KAFKA_VERSION="2.8.1"
```
whitfiea: We need this version rather than the default of 3.3.1? Do we need to set this on deprovision?

image/cli/mascli/functions/gitops_init_kafka
```diff
@@ -239,31 +239,36 @@ function gitops_init_kafka() {
...
-
-    
     #kafka aws provider role specific env variables
     export MAS_CONFIG_DIR=$TEMP_DIR
     export KAFKA_ACTION="install"
-    export KAFKA_NAMESPACE="msk-${MAS_INSTANCE_ID}"  
+    
+    export MAS_INSTANCE_ID=${MAS_INSTANCE_ID:-CLUSTER_ID}
+    export KAFKA_NAMESPACE="msk-${CLUSTER_ID}"
     export KAFKA_VERSION="2.8.1"
```
whitfiea: This wasn't changed in this ver, but i was wondering why 2.8.1?

tekton/src/pipelines/gitops/deprovision-mas-instance.yml.j2
```diff
@@ -186,3 +192,91 @@ spec:
       taskRef:
         kind: Task
         name: gitops-process-mongo-user
+{% if deprovision_manage_addons == true %}
+    - name: gitops-deprovision-efs
```
whitfiea, line 195: gitops-deprovision-efs hould be in the deprovison-cluster pipeline I would have thought? We don't want to remove EFS from the cluster that any existing MAS instances might stil be using.

tekton/src/pipelines/gitops/deprovision-mas-instance.yml.j2
```diff
@@ -186,3 +192,91 @@ spec:
...
+          value: $(params.mas_instance_id)
+        - name: cloud_provider
+          value: $(params.cloud_provider)
+      workspaces:
+        - name: configs
+          workspace: configs
+      taskRef:
+        kind: Task
+        name: gitops-deprovision-efs
+    - name: gitops-delete-kafka-config
```
whitfiea, line 217: I believe for the deprovision of kafka-config we should just call it anyway regardless as if there is no config then it shouldn't cause a problem. So we should remove the deprovision_manage_addons check in the templating.

tekton/src/pipelines/gitops/deprovision-mas-instance.yml.j2
```diff
@@ -186,3 +192,91 @@ spec:
...
+          value: $(params.avp_aws_secret_key)
+        - name: avp_aws_access_key
+          value: $(params.avp_aws_access_key)
+      workspaces:
+        - name: configs
+          workspace: configs
+      taskRef:
+        kind: Task
+        name: gitops-delete-kafka-config
+    - name: gitops-deprovision-kafka
```
whitfiea: There is one kafka per cluster so gitops-deprovision-kafka needs to be in the deprovision-cluster pipeline

tekton/src/pipelines/gitops/provision-mas-instance.yml.j2
```diff
@@ -255,3 +261,91 @@ spec:
       workspaces:
         - name: configs
           workspace: configs
+{% if provision_manage_addons == true %}
```
whitfiea: The condition here should be a `when` condition rather than a templating condition. This would then when condition check is done each time the pipeline runs rather than when the pipeline resource is created. the condition would be something like the following for the kafka and kafka-config task:

```
      when:
      - input: "$(params.mas_app_channel_iot)"
        operator: notin
        values: [""]
```

I think for efs we should just install it as mulitple apps and fvt itself need efs.

whitfiea (changes requested):

image/cli/mascli/functions/gitops_deprovision_kafka
padmankosalaram: KAFKA_VERSION variable removed as its not required in uninstall flow

image/cli/mascli/functions/gitops_init_kafka
padmankosalaram: Added KAFKA_VERSION as parameter and removed the hard coding. We can now set it in pipeline run

tekton/src/pipelines/gitops/provision-mas-instance.yml.j2
padmankosalaram: added the conditional logic to  cli/tekton/src/pipelines/gitops/provision-mas-instance.yml.j2

tekton/src/pipelines/gitops/deprovision-mas-instance.yml.j2
padmankosalaram: moved the kafka uninstall to deprovision-cluster.yml.j2

tekton/src/pipelines/gitops/deprovision-mas-instance.yml.j2
padmankosalaram: removed the deprovision_manage_addons from deprovision-mas-instance.yml.j2 and  generate-tekton-pipelines.yml

tekton/src/pipelines/gitops/deprovision-mas-instance.yml.j2
padmankosalaram: gitops-provision-efs creates a configurable storage for each mas instance. 
similarly gitops-deprovision-efs deletes the configurable storage at mas instance level.  
Hence include deprovision efs is included in deprovision-mas-instance.yml.j2

whitfiea: It is an EFS per MAS instance including the storage class, my mistake.

whitfiea (approved):

PR #460, opened 2023-09-06
[minor] Support option to delete mirror registry

caroazad:
Add support to delete the mirror registry


caroazad:
Test was successful. Is there anything we would like to change? Here is a screenshot after the deletion of the registry
Screenshot 2023-09-07 at 4 02 47 PM

docs/commands/teardown-registry.md
```diff
@@ -0,0 +1,9 @@
+Tear Down Private Registry
+===============================================================================
+
```
whitfiea, line 4: Can we add some more words to this file. What does it do? Why would you use it? Any impacts of doing this? Will running this cause my MAS system to stop working? Do I have to mirror again?

whitfiea (changes requested):
Comment on the readme, and need to remove the image/cli/install-ansible/ibm-mas_devops.tar.gz file

whitfiea (approved):

PR #473, opened 2023-09-11
[patch] minor changes - path update 

padmankosalaram:
- [ ] update to dir path /tmp
- [ ] oc login check and exit 1 or 0 based on provision / deprovision flow
- [ ] vpc check exit 0 in deprovision flow

### DIFF
--- a/tekton/src/tasks/gitops/gitops-deprovision-suite-config.yml.j2
+++ b/tekton/src/tasks/gitops/gitops-deprovision-suite-config.yml.j2
@@ -58,7 +58,7 @@ spec:
   steps:
     - args:
       - |-
-        mkdir -p /workspace/configs/$GITHUB_REPO/deprovision-suite-config
+        mkdir -p /tmp/deprovision-suite-config
 
         git config --global user.name "yourUsername"
         git config --global user.email "you@example.com"
@@ -69,7 +69,11 @@ spec:
         export OCP_USERNAME=$(cat $ROSA_CONFIG | yq -r '.data.username')
         export ROSA_CLUSTER_ADMIN_PASSWORD=$(cat $ROSA_CONFIG | yq -r '.data.admin_password')
         oc login -u $OCP_USERNAME -p $ROSA_CLUSTER_ADMIN_PASSWORD $OCP_SERVER --insecure-skip-tls-verify
-
+        OC_RC=$?
+        if [ $OC_RC -ne 0 ]; then
+          echo "oc login failed, exiting with error.."
+          exit 1
+        fi
         echo "Set k8s context"
         export K8S_AUTH_CONTEXT=$(oc whoami -c)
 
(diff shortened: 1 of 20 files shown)


tekton/src/tasks/gitops/gitops-deprovision-suite-config.yml.j2
```diff
@@ -69,7 +69,11 @@ spec:
         export OCP_USERNAME=$(cat $ROSA_CONFIG | yq -r '.data.username')
         export ROSA_CLUSTER_ADMIN_PASSWORD=$(cat $ROSA_CONFIG | yq -r '.data.admin_password')
         oc login -u $OCP_USERNAME -p $ROSA_CLUSTER_ADMIN_PASSWORD $OCP_SERVER --insecure-skip-tls-verify
-
+        OC_RC=$?
+        if [ $OC_RC -ne 0 ]; then
+          echo "oc login failed, exiting with error.."
+          exit 1
+        fi
```
whitfiea, line 76: I think we have to be careful here. The cluster might not exist but we still want to carry on and remove the pieces from github. This then links to the other piece of work which is to allow deprovsion to run so we need to timeout the argo sync check. So I think we should log the error here but carry on.

tekton/src/tasks/gitops/gitops-deprovision-suite-workspace.yml.j2
```diff
@@ -66,7 +66,11 @@ spec:
         export OCP_USERNAME=$(cat $ROSA_CONFIG | yq -r '.data.username')
         export ROSA_CLUSTER_ADMIN_PASSWORD=$(cat $ROSA_CONFIG | yq -r '.data.admin_password')
         oc login -u $OCP_USERNAME -p $ROSA_CLUSTER_ADMIN_PASSWORD $OCP_SERVER --insecure-skip-tls-verify
-
+        OC_RC=$?
+        if [ $OC_RC -ne 0 ]; then
+          echo "oc login failed, exiting with error.."
+          exit 1
+        fi
```
whitfiea, line 73: I think we have to be careful here. The cluster might not exist but we still want to carry on and remove the pieces from github. This then links to the other piece of work which is to allow deprovsion to run so we need to timeout the argo sync check. So I think we should log the error here but carry on.

whitfiea (changes requested):

tekton/src/tasks/gitops/gitops-deprovision-suite-config.yml.j2
padmankosalaram: added the logic to exit with 0 in deprovision flow

tekton/src/tasks/gitops/gitops-deprovision-suite-workspace.yml.j2
padmankosalaram: added the logic to exit with 0 in deprovision flow

whitfiea (approved):

PR #480, opened 2023-09-12
[patch] add deprovision support for smtp and idp

padmankosalaram:
Deprovision smtpcfg and idpcfg
https://jsw.ibm.com/browse/MASCORE-241?filter=-1


whitfiea (changes requested):
Why do we need to pass in all the smtp and ldap options on the deprovsion? If we are just removing the files from github and the secret then we don't need any of those details do we?

whitfiea (approved):

PR #482, opened 2023-09-12
[patch] fix kafka byo

andrercm:
Changes the kafka taskdef to only skip kafka task if iot is not being installed or byo is not defined


whitfiea (changes requested):
What about if kafka action is uninmtall?

andrercm:
> What about if kafka action is uninmtall?

@whitfiea i think instead of doing the change in this PR, I'll have to change fvts travis.yaml to set `kafka_action_system: install` just like we have for db2 as in: https://github.ibm.com/maximoappsuite/fvt-dev/blob/fvtcpd/.travis.yml#L54

What you think?

andrercm:
This PR would just be a temp fix in fact to not break fvts as of today... but i think at this point it's just better to add the new var across all fvt's.

PR #483, opened 2023-09-13
[patch]Add the tasks/pipeline to deprovsion a DB2 instance using gitops

Yolandabjyf:
For story [Deprovision DB2 instance pipeline tasks](https://jsw.ibm.com/browse/MASCORE-223)


tekton/src/tasks/gitops/gitops-deprovsion-db2u.yml.j2
```diff
@@ -0,0 +1,99 @@
...
+
+        if [ -z $ARGOCD_URL ] || [ -z $ARGOCD_USERNAME ] || [ -z $ARGOCD_PASSWORD ]; then
+          export ARGOCD_URL=$(oc get route  openshift-gitops-server -n openshift-gitops -ojsonpath='{.spec.host}' --context $K8S_AUTH_CONTEXT)
+          export ARGOCD_USERNAME=admin
+          export ARGOCD_PASSWORD=$(oc get secret openshift-gitops-cluster -n openshift-gitops -ojsonpath='{.data.admin\.password}' --context $K8S_AUTH_CONTEXT | base64 -d ; echo)
+        fi
+      
+        mkdir -p /workspace/configs/$GITHUB_REPO
+        mas gitops-deprovision-db2u -a $ACCOUNT -c $CLUSTER_NAME -m $MAS_INSTANCE_ID \
+          --dir /workspace/configs/$GITHUB_REPO \
```
whitfiea: There is a recent change to not use the shared /workspace/configs for the git dir. This is because we end up with tasks editing the same folder when run at the same time. See https://github.com/ibm-mas/cli/blob/gitops/tekton/src/tasks/gitops/gitops-deprovision-suite-config.yml.j2#L88 for the new type of /tmp location we need to set (and make sure is created)

tekton/src/tasks/gitops/gitops-deprovsion-db2u.yml.j2
```diff
@@ -0,0 +1,99 @@
...
+      - |-
+        git config --global user.name "yourUsername"
+        git config --global user.email "you@example.com"
+        git config --global user.password $GITHUB_PAT
+
+        ROSA_CONFIG=/workspace/configs/tmp-rosa-init/rosa-$(params.cluster_name)-details.yaml
+        export OCP_SERVER=$(cat $ROSA_CONFIG | yq -r '.data.api_url')
+        export OCP_USERNAME=$(cat $ROSA_CONFIG | yq -r '.data.username')
+        export ROSA_CLUSTER_ADMIN_PASSWORD=$(cat $ROSA_CONFIG | yq -r '.data.admin_password')
+        oc login -u $OCP_USERNAME -p $ROSA_CLUSTER_ADMIN_PASSWORD $OCP_SERVER --insecure-skip-tls-verify
```
whitfiea, line 70: Can you add the check like this https://github.com/ibm-mas/cli/blob/gitops/tekton/src/tasks/gitops/gitops-deprovision-suite-config.yml.j2#L73 to say if there is no oc cluster then just exit as there is nothing to remove.

tekton/src/pipelines/gitops/deprovision-mas-instance.yml.j2
```diff
@@ -192,6 +196,46 @@ spec:
       taskRef:
         kind: Task
         name: gitops-process-mongo-user
+    - name: gitops-deprovsion-db2u
+      runAfter:
+        - gitops-process-mongo-user
```
whitfiea: Does Db2 need EFS? If so then we probably need to run it before the deprovison-efs which is currently going to happen at the same time. I believe we can remove db2 at the same time as mongo user, so just change this gitops-deprovsion-db2u to runAfter: gitops-deprovision-suite

whitfiea (changes requested):

whitfiea (approved):

PR #491, opened 2023-09-15
[minor] Add Health FVT pipeline

karlinglee93:
Add the Health pipeline to master branch.
Changes have been tested fully on my personal fvt cluster - healthjn.
截屏2023-09-15 上午9 36 59


whitfiea (changes requested):
@karlinglee93 Can you confirm that the tests pass as well? Rather than just the test task runs?

karlinglee93:
> @karlinglee93 Can you confirm that the tests pass as well? Rather than just the test task runs?

Hi @whitfiea , I cannot guarantee every test would pass, because there are around 600 test cases. And We keep monitoring and fixing the defects every day and the pass rate right now is around 70% (Our best result is over 95% before). We prefer to onboard the pipeline first, and keep improving the pass rate. Do you agree?

whitfiea (approved):
Test code changed so it now only contains passing testcases

PR #504, opened 2023-09-19
[minor] Require --secret-data flag to include secrets in must-gather

mattlrx:
some secret data are sensitive like super user name and password, the collection of the secret has been changed to do a oc describe secret by default and oc get secret when the --secret-data is specified.


andrercm:
@mattlrx has this change been tested somewhere to ensure everything works as expected and no regressions?

mattlrx:
@andrercm 
I tested the change with the cli image built locally and against my cluster, ran a must gather with and without the option and verified the secrets were collected with and without the data.

andrercm:
@mattlrx what's the background behind this change? Is this following a compliant reasoning thinking on the security perspective? Have we ran this past someone from security team? If I understood correctly we're enabling an option to retrieve secret's content along with the must-gather logs? Wouldn't we be potentially infringing sensitive data constraints if we collected secret's data?

mattlrx:
@andrercm 

it is actually the oposite, currently, we are collecting the secrets using "oc get secret -o yaml", this includes the data of the secret, this can be decoded.

the change i propose is to make the default collection of the secret with "oc describe secret", which only collects the name of the secret key with their size

and if needed, collect the secret data by specifying a flag --secret-data in the must-gather command.

durera (approved):

PR #512, opened 2023-09-21
 [patch] [MASCORE-309] fvt pipeline support 

padmankosalaram:


tekton/src/tasks/gitops/gitops-provision-mas-fvt-core.yml.j2
```diff
@@ -0,0 +1,120 @@
...
+        fi
+
+        echo "Set k8s context"
+        export K8S_AUTH_CONTEXT=$(oc whoami -c)
+
+        export MAS_CONFIG_DIR=/tmp/additional-configs
+        mkdir -p $MAS_CONFIG_DIR
+
+        if [[ -e /workspace/shared-entitlement/entitlement.lic ]]
+          cat /workspace/shared-entitlement/entitlement.lic > /tmp/additional-configs/entitlement.lic
```
whitfiea, line 90: Can we just cp all files under /workspace/shared-entitlement and /workspace/shared-additional-configs to the working dir /tmp/additional-configs ? this would then allow for new configs to be added and not have to worry about changing this.

whitfiea (changes requested):
Need to remove the ibm-mas_fvt.tar.gz as well

PR #524, opened 2023-09-27
[patch] Fix SLS 3.8.x registry mismatch

rawa-resul:
## Description

SLS `3.8.0` changed the registry for api-licensing from `cp.icr.io/cp` to `icr.io/cpopen`. For fresh installations this works fine as ansible-devops has been updated. However, we have noticed that during `mas update` when updating from a previous version of SLS to `3.8.0` the registry would stay as `cp.icr.io/cp` and cause the now pod to not pull the image.

- Added new task sls-registry-update to run when the latest catalog is chosen to update all LicenseService CRs' registries to `icr.io/cpopen`

Screenshot 2023-09-28 at 12 25 26 am

### DIFF
--- a/image/cli/mascli/functions/update
+++ b/image/cli/mascli/functions/update
@@ -112,6 +112,11 @@ function update() {
   echo_h4 "IBM Operator Catalog" "    "
   echo_reset_dim "Catalog Version ..................... ${COLOR_MAGENTA}${MAS_CATALOG_VERSION}"
 
+  # SLS changed the registry for api-licensing in 3.8.0
+  if [[ "$MAS_CATALOG_VERSION" == "v8-230926-amd64" ]]; then
+    SLS_ICR_CPOPEN=icr.io/cpopen
+  fi
+
   echo
   reset_colors
   if [[ "$NO_CONFIRM" != "true" ]]; then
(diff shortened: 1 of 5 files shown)


image/cli/mascli/functions/update
```diff
@@ -112,6 +112,15 @@ function update() {
   echo_h4 "IBM Operator Catalog" "    "
   echo_reset_dim "Catalog Version ..................... ${COLOR_MAGENTA}${MAS_CATALOG_VERSION}"
 
+  # In SLS 3.8.0 registry was changed to icr.io/cpopen, however
+  # when updating from a previous version it would not automatically
+  # change the value from cp.icr.io/cp hence we need to patch it
+  if [[ "$MAS_CATALOG_VERSION" == "v8-230926-amd64" ]]; then
```
whitfiea, line 118: is there harm in not having this check and just doing the update regardless as to if it is correct or not? There is the case of someone doing the update on the next catalog after this but not this one (so skipping this one).

whitfiea (changes requested):

image/cli/mascli/functions/update
rawa-resul: if this update happens on previous catalogs it will break SLS. I am going to release SLS 3.8.1 soon which will autofix this issue. So really we just need to do it for any catalog that has 3.8.0 in it which currently is `v8-230926-amd64`

whitfiea (approved):

PR #568, opened 2023-10-19
[patch] Fixes for #302, #527 and #605

### LINKED ISSUE #302: Add option to select the name of the ingress secret
By default `mas install` assumes that the Openshift cluster has the default ingress certificates with a secret name of: `router-certs-default`.

But we have replaced the default ingress certificates following this documentation:
https://docs.openshift.com/container-platform/4.10/security/certificates/replacing-default-ingress-certificate.html

We would need to be able to set the name of the secret, since the documentation allows you to set any name you want.

Thanks!

### LINKED ISSUE #527: Non-interactive mas install is interactive instead - at least when you define additional configurations
I have tried the non-interactive mode of mas install cli by running this command

mas install \
  -c v8-amd64 \
  --ibm-entitlement-key myentitlementkey \
  --license-id 0242ac110002 \
  --license-file /tmp/entitlement.lic \
  --uds-email mas@test.com \
  --uds-firstname mas \
  --uds-lastname mas -i inst1 -w main -W "Main Workspace" \
  --additional-configs /tmp/configs/ \
  --mas-channel 8.11.x \
  --manage-channel 8.7.x \
  --manage-jdbc workspace-application \
  --manage-server-bundle-size dev \
  --manage-components "base=latest" \
  --storage-rwo ocs-storagecluster-ceph-rbd \
  --storage-rwx ocs-storagecluster-cephfs \
  --storage-pipeline ocs-storagecluster-cephfs \
  --storage-accessmode ReadWriteMany \
  --non-prod --no-wait-for-pvcs

I would expect not having to interact after that, but following this command I got this asking for my input

image

I was hoping to build a wrapper on top of this mas install cli, running it in non-interactive mode through a jenkins pipeline/job, but this request for input, pretty much prevent this to be done. Please make the non-interactive mode really non-interactive.

### LINKED ISSUE #605: Installing IoT with CLI 7.6.0 installer fails with undefined properties
Installing 8.11.1 with latest CLI installer, fails at app-install-iot. mas_app_settings_iot_fpl_pvc_storage_class property must be defined when IoT deployment size = 'small'.

andrercm:
Fixes for:

- Fixed `OCP_CONSOLE_ROUTE` undefined when running non-interactive mode (with/out --no-confirm)
- Fixed custom storage classes not setting the iot storage classes in the cli pipeline, now we'll use common rwo storage used by pipeline to set IoT storage classes (since user is not prompted to pick iot storage class)
- https://github.com/ibm-mas/cli/issues/527
- https://github.com/ibm-mas/cli/issues/302#issuecomment-1789888186
- https://github.com/ibm-mas/cli/issues/605


image/cli/mascli/functions/pipeline_config_advanced
```diff
@@ -48,7 +48,10 @@ function config_pipeline_advanced() {
       prompt_for_input "+ MongoDB Namespace" MONGODB_NAMESPACE "mongoce"
     fi
     echo
-
+    if [[ -z "$OCP_INGRESS_TLS_SECRET_NAME" ]]; then
+      OCP_INGRESS_TLS_SECRET_NAME="router-certs-default"
```
durera: I think the default here should still be "" ... most users - even when doing the advanced settings don't want to have to set this secret.

Probably have a: `Override the ingress certificate lookup? y/N` prompt .. so that most people can just click though and retain the same logic we have today.

tekton/src/pipelines/taskdefs/cluster-setup/ocp-verify.yml.j2
```diff
@@ -17,3 +17,5 @@
       value: {{ verify_workloads | default("True") }}
     - name: verify_ingress
       value: {{ verify_ingress | default("True") }}
+    - name: ocp_ingress_tls_secret_name
```
durera, line 20: Shouldn't this have a default value of "", rather than being a required parameter?

tekton/src/pipelines/rollback.yml.j2
```diff
@@ -21,6 +21,11 @@ spec:
       type: string
 {% endif %}
 
+    # Cluster config
```
durera, line 24: Did mean to add this here?  Isn't this needed only for install?

durera (changes requested):

image/cli/mascli/functions/pipeline_config_advanced
andrercm: Ok, I'll change the logic to ask if user really wants to override this prior prompting them to type the cert secret.
Something as:
```
    if prompt_for_confirm "Change default cluster ingress certificate secret?"; then
        prompt_for_input "+ Cluster ingress certificate secret" OCP_INGRESS_TLS_SECRET_NAME $OCP_INGRESS_TLS_SECRET_NAME
    fi
```

tekton/src/pipelines/taskdefs/cluster-setup/ocp-verify.yml.j2
andrercm: Hmmn.. I thought that since this is coming from `params`, the default to "" would be coming from here? 
https://github.com/ibm-mas/cli/blob/andrercm-multiple-fixes/tekton/src/params/install.yml.j2#L4

tekton/src/pipelines/rollback.yml.j2
andrercm: This became a requirement because `rollback` pipeline makes use of the `ocp-verify` taskdef:
https://github.com/ibm-mas/cli/blob/andrercm-multiple-fixes/tekton/src/pipelines/rollback.yml.j2#L71

and `ocp-verify` taskdef sets the `ocp_ingress_tls_secret_name` param after my changes:
https://github.com/ibm-mas/cli/blob/andrercm-multiple-fixes/tekton/src/pipelines/taskdefs/cluster-setup/ocp-verify.yml.j2#L20

I guess if we don't plan to make use of that property in other pipelines, i could remove that from the taskdef and just add it as a single entry in the `install` pipeline? Somewhere over here:
https://github.com/ibm-mas/cli/blob/andrercm-multiple-fixes/tekton/src/pipelines/install.yml.j2#L45

image/cli/mascli/functions/pipeline_config_advanced
andrercm: image

tekton/src/pipelines/taskdefs/cluster-setup/ocp-verify.yml.j2
andrercm: image

durera (approved):

durera:
Merged, but we should avoid doing multiple fixes in a single PR where possible.

PR #595, opened 2023-11-02
[minor] Allow Mustgather non-interactive to connect to remote OCP Cluster

mattlrx:
when running the must gather in non insteractive mode, there is no connection to OCP established and the must gather does not happen, 
added parameters to the must gather function to specify the server and the token, set them as global variable to reuse in the connect function.


image/cli/mascli/functions/must_gather
```diff
@@ -93,6 +108,13 @@ function mustgather() {
 
   TIMESTAMP=$(date +"%Y%m%d-%H%M%S")
 
+  # use the provided login details to authenticate against OCP
+  export OCP_LOGIN_URL=$OCP_LOGIN_URL
+  export OCP_LOGIN_TOKEN=$OCP_LOGIN_TOKEN
+  export OCP_DISABLE_TLS_VERIFY=$OCP_DISABLE_TLS_VERIFY
+
+  connect
```
whitfiea, line 116: If we are running mustgather in the cluster then it will already have a connection, but the call to `connect` here will mean that it will prompted the user to confirm. Can we change this so we only connect if we have the OCP_LOGIN_URL and OCP_LOGIN_TOKEN set?

whitfiea (changes requested):

durera (changes requested):
All the non-interactive modes work on the assumption you are already connected to the cluster.  Changing this just for the must-gather doesn't make sense, would make it inconsistent with all the other CLI functions.

Even the interactive mode is designed on the basis you are already connected, and assumes you want to use the cluster you are already connected to, it only prompts for connection information if you are not connected to something, or if you say "no, I didn't mean to use this cluster" when prompted to confirm the cluster details.

I don't think we should add this at all personally, it's unnecessary noise/complexity in the param space because if you are doing this non-interactive you can just put `oc login` before the command .. just like if I want to run `oc get pods` there isn't an option to specifc the login server details, I have to `oc login` first.  

But if we are going to add it, then it needs to be added universally as a new set of flags that are the same on all functions and work for all functions that expect to be connected to a cluster to work, so that the behaviour across the CLI is consistent.  So I would suggest that they need to be the long form only so as not to reserve any short flags.

e.g. mirror-images doesn't require an OCP connection, but install, update, upgrade, rollback, uninstall, configure-airgap etc all do.

mattlrx:
thanks @durera for the feedback, I did not think of that.
I can actually run it like this:
podman run  -v /root/images:/mnt/registry:z quay.io/ibmmas/cli /bin/bash -c "oc login --token=sha256~XFnSk...fc8U --server=https://api.elm-support-ocp.os.fyre.ibm.com:6443 --insecure-skip-tls-verify; mas must-gather"

we can cancel the pull request and close it

mattlrx:
change is not required since we can run the must gather in non-interactive mode when passing the oc login command before the mas must-gather command:
podman run -v /root/images:/mnt/registry:z quay.io/ibmmas/cli /bin/bash -c "oc login --token=sha256~XFnSk...fc8U --server=https://api.elm-support-ocp.os.fyre.ibm.com:6443/ --insecure-skip-tls-verify; mas must-gather"

PR #608, opened 2023-11-10
[patch] adding default params 

padmankosalaram:


image/cli/mascli/templates/gitops/ibm-mas-core-suite-app-config.yaml.j2
```diff
@@ -23,7 +23,7 @@ spec:
             mas_instance_id: {{ MAS_INSTANCE_ID }}
             mas_app_id: {{ MAS_APP_ID }}
             mas_app_namespace: {{ MAS_APP_NAMESPACE }}
-
+            mas_app_settings_server_bundles_size: {{ MAS_APP_SETTINGS_SERVER_BUNDLES_SIZE }}
```
whitfiea, line 27: This would only be valid for manage, so we need to add a condition on it like the others in this file

whitfiea (changes requested):

whitfiea (approved):

PR #610, opened 2023-11-10
[minor] Enable ipv6 testing internally for Fyre

xinyuwu1:
This PR added the ability to run the fvt pipeline with the enable_ipv6 flag.  When enable_ipv6 is set to true, the pipeline can provision an internal fyre cluster with IPv6 enabled for testing.

Tested with pfvt for both enable_ipv6 true on FYRE and false (by default) on ibmcloud. Provisioned and pipeline started running correctly


Makefile
```diff
@@ -6,7 +6,7 @@
 
 ansible-build:
 	ansible-galaxy collection build --output-path image/cli/install-ansible ../ansible-devops/ibm/mas_devops --force
-	mv image/cli/install-ansible/ibm-mas_devops-100.0.0.tar.gz image/cli/install-ansible/ibm-mas_devops.tar.gz
+	mv /root/github/ibm-mas/ansible-devops/ibm-mas_devops-100.0.0.tar.gz image/cli/install-ansible/ibm-mas_devops.tar.gz
```
whitfiea: I don't think we want this change here

whitfiea (changes requested):

Makefile
xinyuwu1: Sorry I thought I pushed this but it was still sitting in my local. Please review again. Thanks

whitfiea (approved):

PR #615, opened 2023-11-12
[minor] Overhaul FVT pipeline integration

durera:
Start to use pipelines in pipelines to begin to bring install / install-with-fvt back closer to one another.  Install with FVT pipeline will have a few additional "kick off" tasks that launch completely separate FVT pipelines.

Also:
- Simplify assist pipeline build (remove unnecessary templating)
- Introduce dedicated Core FVT pipeline
- Introduce dedicated Assist FVT pipeline
- Introduce dedicated Manage, Manage IS, Manage Mobile, and Manage Regression FVT pipelines
- Update [build-tekton.sh](https://github.com/ibm-mas/cli/pull/615/files#diff-5beae48ecd6e0eb3ce35d3bcf8b68c9a458f07bc305d6f5dd6f10388e403be92) to support build of tasks or pipelines only using $1 arg set to `tasks` or `pipelines`

image
image
image
image


alequint:
@durera it may be better if we could merge this PR https://github.com/ibm-mas/cli/pull/614 before this one. It is a small quick fix for a problem introduced when cv-app tests were onboarded, related to the different run-after of ivtcore and, now, cv, when running in install-with-fvt and when running in fvt-manage.

this is more to get things right from start in this reorg.... or is too late? :)

durera:
> @durera it may be better if we could merge this PR #614 before this one. It is a small quick fix for a problem introduced when cv-app tests were onboarded, related to the different run-after of ivtcore and, now, cv, when running in install-with-fvt and when running in fvt-manage.
> 
> this is more to get things right from start in this reorg.... or is too late? :)

I would just close that PR, and take a look whether the same problem exists in the updated manage pipelines here as it's not 1:1 ... eg. manage-regr pipeline is **just** the manage regression tasks for example.

PR #618, opened 2023-11-13
[patch] New task definition for mobile FVT

gabrielbonamicoibm:
- New task for mobile tests;
- fix for civil defects android test.


leo-miran (dismissed):

tekton/src/tasks/fvt/fvt-mobile.yml.j2
```diff
@@ -0,0 +1,159 @@
...
+      type: string
+      description: Instance ID of the target test environment
+    - name: mas_workspace_id
+      type: string
+      description: Workspace ID in MAS to use for running the tests
+      default: "masdev"
+
+    # Test-specific Information (all optional)
+    # -------------------------------------------------------------------------
+    - name: ddp_apikey
```
durera: Do you actually use ddp_apikey, partium_username, partium_password in the test?  I thought these were only used in a handful of Core FVTs?

leo-miran: We do not, but these were added a while ago in all tasks so I figured this was needed. I would be more than glad to remove these. Should I?

leo-miran: I have just removed these params. Thanks!

leo-miran (approved):

alequint (approved):

durera (approved):

PR #626, opened 2023-11-15
Add DRO and DRO config to suite-configs for gitops

whitfiea:
Adds DRO gitops app that will install DRO at the cluster level. The apikey we use is stored in secretsmanager. The DRO app also needs an entitlement key but just an entitlement key rather than a pull secret (even though the secret key is called PULL_SECRET!).

Also sets up the bascfg to use this DRO service rather than use suds. The host and cert are computed from the OCP cluster we are connected to, so the assumption (in the pipeline task) is that the DRO instance is on the same cluster that the DRO config (bascfg) is being setup for. The gitops function itself can take a host and ca_cert if needed


padmankosalaram (approved):
Please merge the changes

PR #639, opened 2023-11-17
Update the name for apps to remove core-suite and add workspace

whitfiea:
Adds the workspace to the name of the argocd app so we can potentially have mulitple app configs in seperate workspaces. Also removes the core-suite part of the names as the apps are not really part of core and it was making the file name longer.


padmankosalaram (approved):
Looks good, please merge

PR #662, opened 2023-11-28
Alow db2ucluster cr settings to be provided

whitfiea:
Allows the db2ucluster settings to be passed in as yaml files so we can set it for manage. Also calls the db2 commands rather than calling the db2_setup_for_manage role.

Also adds a few fixes etc for running the app pipelines


padmankosalaram (approved):
Changes looks good. Please merge

PR #670, opened 2023-11-29
[patch] fvt mobile updates

leo-miran:
- Updates to mobile fvt pipeline and tasks
- Setting defaults to products launchfvt params to false in install_with_fvt definition


durera (changes requested):
As discussed in Slack

leo-miran:
Will raise a new pr with the comments from David as well as new additions

PR #676, opened 2023-12-01
[minor] mobile pipeline and tasks params update

leo-miran:
- Updating fvt-mobile pipeline params
- fvt-mobile pipeline tasks reorder
- Setting defult to false to all launchfvt_xxx params in install-with-fvt pipeline definition


alequint (dismissed):
@leo-miran this is actually a `minor`update (not a `patch`) as you are adding parameters and vars here and there

leo-miran:
> @leo-miran this is actually a `minor`update (not a `patch`) as you are adding parameters and vars here and there

I could not find a CHANGES.md file to update with the minor changes details. So I just updated the PR title.

durera (approved):

PR #686, opened 2023-12-11
 [patch] [mascore-1112] Apply db2cfg settings only when required

padmankosalaram:


image/cli/mascli/functions/gitops_db2u_database
```diff
@@ -576,19 +592,43 @@ DB2_WORKLOAD: '${DB2_WORKLOAD}'"
...
     [ $rc -ne 0 ] && exit $rc
+
+
+    if [[ "${DB2CFG_SETTINGS_APPLIED}" != "true" ]]; then
+      echo "sm_update_secret ...DB2CFG_SETTINGS_APPLIED with true"
+      export DB2CFG_SETTINGS_APPLIED="true"
+      sm_update_secret ${DB2CFG_SETTINGS_APPLIED_SECRET} "{ \"db2cfg_settings_applied\": \"${DB2CFG_SETTINGS_APPLIED}\" }"
+
+      echo
+      echo_reset_dim "- Re-generating DB2U Database file ${GITOPS_APPS_DIR}/${MAS_INSTANCE_ID}.${MAS_APP_ID}.ibm-db2u-database.yaml"
```
whitfiea: As we are updating a secret in secrets manager and using that to determine whether the settings have been applied or not then I think we should remove the change to "Re-generating DB2U Database file" as we don't need it in two places. Either we put it in the secretsmanager or we put it in the helm values file but not both.

whitfiea (changes requested):

padmankosalaram:
Created another PR https://github.com/ibm-mas/cli/pull/688, closing this one.

PR #710, opened 2024-01-02
[patch] minor changes

padmankosalaram:


tekton/src/tasks/gitops/gitops-deprovision-cos.yml.j2
```diff
@@ -89,8 +89,8 @@ spec:
         oc login -u $OCP_USERNAME -p $ROSA_CLUSTER_ADMIN_PASSWORD $OCP_SERVER --insecure-skip-tls-verify
         OC_RC=$?
         if [ $OC_RC -ne 0 ]; then
-          echo "oc login failed"
-          exit 1
+          echo "oc login failed, so assume there is no cos to deprovision, exiting.."
```
whitfiea: The current deprovision-cos doesn't actually need to connect to the OCP cluster, so I think rather than return with an exit of 0 we should just remove the `oc login` completely.

whitfiea (changes requested):

whitfiea (approved):

PR #714, opened 2024-01-03
[patch] predict changes

padmankosalaram:


tekton/src/pipelines/gitops/provision-mas-apps.yml.j2
```diff
@@ -869,3 +909,264 @@ spec:
...
+        - name: shared-gitops-configs
+          workspace: shared-gitops-configs
+      taskRef:
+        kind: Task
+        name: gitops-cp4d-service
+      when:
+        - input: "$(params.mas_app_channel_predict)"
+          operator: notin
+          values: [""]
+        - input: "$(params.cp4d_watson_machine_learning_action)"
```
whitfiea, line 1086: I think we are missing cp4d_watson_machine_learning_action (and the other cp4d actions other than wsl) from that params

whitfiea (changes requested):

PR #726, opened 2024-01-10
[patch] Preparation for UDS to DRO migration in Feb update

lokesh-sreedhara:
MAS migration to DRO
- migration of DRO will be handled as part of `mas update` only when catalog is >= Jan
- migration of DRO occurs only if certain conditions are met. 
     -  A valid UDS needs to be running on the cluster
     - User needs to confirm yes to the dro migration question 
     - if user enters yes, the DRO_MIGRATION flag is set to true
     - all available Suite CR are verifies for Versions, DRO is not supported in MAS 8.9 
     - all avalaible bascfg is checked to ensure its using a local instance of UDS, remote UDS is not supported for migration.
     - if all conditions are met, teckton pipeline for update is called and DRO_MIGRATION flag is passed on to the UDS taskRuns
     
    CLI screen shot:
    
image

ansible-devops related changes are here: https://github.com/ibm-mas/ansible-devops/pull/1165


andrercm (dismissed):
LGTM - as per discussion with david, we just need to ensure that this is only effective when Feb catalog is released.

durera (approved):

PR #739, opened 2024-01-19
[minor] Ancestors and Automation Scripts Tests

chiragcsatani:
**PYTEST:**

**For Ancestors feature:**
Added below scripts for asset and workorder object to test ancestors.

test_workorder_ancestors_creation
test_asset_ancestors_creation

**For AutomationScripts:**
Added below scripts for asset and workorder object using automation scripts:

test_automationscript_wo_creation
test_automationscript_wo_deletion
test_automationscript_asset_creation
test_automationscript_asset_deletion

All these tests are tested in local as well in fyre pfvt clusters and working okay.

Updated phase1.yml.j2 file with these test suites for pytest


tekton/src/pipelines/taskdefs/fvt-manage/phase1.yml.j2
```diff
@@ -32,3 +28,25 @@
...
+- name: fvt-manage-ancestors
+  {{ lookup('template', 'taskdefs/fvt-manage/api/taskref.yml.j2') | indent(2) }}
+  params:
+    {{ lookup('template', 'taskdefs/fvt-manage/api/params.yml.j2') | indent(4) }}
+    - name: fvt_test_suite_prefix
+      value: core
+    - name: fvt_test_suite
+      value: ancestors
+  runAfter:
+    - fvt-manage-bvt
```
alequint: the name of setup task is fvt-manage-setup, even though suite name has changed to be bvt instead. Here we need to stay with that name: https://github.com/ibm-mas/cli/pull/739/files#diff-c0bd12dab72895b7b1363d8040a9c07bad8212f5f54d52105292a88d4756b8ffR10

Also, if ancestors and automationscripts depend on setup they cannot be in phase 1 and should be moved to a later phase (please move to the ones with less tests).

Once they are moved, the runAfter list must also be updated to follow other tests in the same phase as well as the tests in the next phase must consider the two new tests in their own "runAfter" list

alequint (changes requested):

alequint:
Integrated and executed during tests:

image

@chiragcsatani let me know if we can move this PR from draft to Ready for Review and then be merged

alequint (approved):

PR #755, opened 2024-01-29
[minor] Removed metric api

unnati-solanki-git:
**fvt-manage-metrics-api** - This is a selenium framework-based API test that gets the token from the browser cache store and then hits the endpoint to validate the Prometheus metrics. Currently, failing while getting into Manage. 

**fvt-manage-monitoring** - This is an alternative to metric-api test that already validates the same thing directly in the cluster using open shift APIs. 

Hence, this change will no longer keep metric-api in CLI.

image


alequint (approved):

alequint:
Looks good to me! Thanks @unnati-solanki-git

alequint:
Tested resources applying successfully from this branch:

image

Merging now.

PR #761, opened 2024-02-01
[patch] manage component IT should be ICD

chriscochran:
Maximo IT as a Manage component is named 'ICD' and not 'IT'.


andrercm (approved):
LGTM

PR #779, opened 2024-02-09
[patch]: Add support to UI Core cypress tests

cgirani:


alequint:
@cgirani you have the task now we need to call that task from one of the ibm-mas fvt phases here https://github.com/ibm-mas/cli/blob/master/tekton/src/pipelines/fvt-core.yml.j2#L82 and, eventually, have phase dedicated to run ui tests if that is the case. Once task is invoked that, we should test it a personal pipeline to validate before anything.

alequint:
@cgirani what is missing here for the PR be ready for review? Any pending discussion or testing?

PR #786, opened 2024-02-15
[patch] Add Nvidia GPU operator

caroazad:
Add nvidia gpu operator function


padmankosalaram (changes requested):
Do we need to include deprovision logic for nvidia ?
Deprovision logic for db2 can be found here, I believe we can add similar logic for nvidia in the same file
https://github.com/ibm-mas/cli/blob/37897ff5b105e6c7b7a85bb1eb64bffc73a0da0e/image/cli/mascli/functions/gitops_deprovision_cluster#L258

caroazad:
> Do we need to include deprovision logic for nvidia ? Deprovision logic for db2 can be found here, I believe we can add similar logic for nvidia in the same file
> 
> https://github.com/ibm-mas/cli/blob/37897ff5b105e6c7b7a85bb1eb64bffc73a0da0e/image/cli/mascli/functions/gitops_deprovision_cluster#L258

I added it

padmankosalaram (approved):
Please merge the changes

PR #792, opened 2024-02-19
[patch] pytest artifactory issue

unnati-solanki-git:
The Manage pytest logs were not getting uploaded due to missing env ARTIFACTORY_UPLOAD_DIR.

This change should fix this issue.

### DIFF
--- a/tekton/src/tasks/fvt/fvt-manage-pytest.yml.j2
+++ b/tekton/src/tasks/fvt/fvt-manage-pytest.yml.j2
@@ -109,6 +109,13 @@ spec:
 
       - name: ARTIFACTORY_TOKEN
         value: $(params.artifactory_token)
+        
+      - name: ARTIFACTORY_UPLOAD_DIR
+        valueFrom:
+          secretKeyRef:
+            name: mas-devops
+            key: ARTIFACTORY_UPLOAD_DIR
+            optional: true
       
       # Black and white listing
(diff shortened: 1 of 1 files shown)


tekton/src/tasks/fvt/fvt-manage-pytest.yml.j2
```diff
@@ -108,7 +108,17 @@ spec:
         value: "$(params.fvt_enable_debug)"
 
       - name: ARTIFACTORY_TOKEN
-        value: $(params.artifactory_token)
+        valueFrom:
```
alequint: @unnati-solanki-git as you are updating the token here as well, we need also to remove it from the task parameter here https://github.com/ibm-mas/cli/pull/792/files#diff-810eda5557acae5fc32b41243809e5f54e69c2ca5ed9056a26cc80f9552a5f97R64 and make sure the task defintions where this pytest taskref is invoked is no longer sending the parameter to it.

alequint (changes requested):

tekton/src/tasks/fvt/fvt-manage-pytest.yml.j2
unnati-solanki-git: okay let me remove this for now, will refactor it later.

alequint (approved):

PR #801, opened 2024-02-22
[patch] [mascore-1684] Update deprovision tasks pipelines

padmankosalaram:
As part of https://jsw.ibm.com/browse/MASCORE-1684?filter=-1

Pipelines changes expected on SaaS env to 

    deprovision-mas-app
    deprovision-mas-instance
    deprovision-mas-cluster (the cluster level resources only rather than the cluster itself)

This fix is only concerning Esstentials Inspection (MAS core and MVI only) 
So ignored db2, cp4d and kafka from these deprovision tasks for now.

### DIFF
--- a/tekton/src/tasks/gitops/gitops-delete-kafka-config.yml.j2
+++ b/tekton/src/tasks/gitops/gitops-delete-kafka-config.yml.j2
@@ -26,6 +26,9 @@ spec:
       type: string
     - name: avp_aws_secret_region
       type: string
+    - name: cluster_url
+      type: string
+      default: ""
     - name: argocd_url
       type: string
       default: ""
@@ -55,10 +58,18 @@ spec:
         value: $(params.github_pat)
       - name: SM_AWS_REGION
         value: $(params.avp_aws_secret_region)
+      - name: OCP_SERVER
+        value: $(params.cluster_url)
       - name: ARGOCD_URL
         value: $(params.argocd_url)
       - name: ARGOCD_CHECK
         value: $(params.argocd_check)
+    envFrom:
+      - configMapRef:
+          name: environment-properties
+          optional: true
+      - secretRef:
+          name: secure-properties
   steps:
     - args:
       - |-
--- a/tekton/src/tasks/gitops/gitops-deprovision-app-config.yml.j2
+++ b/tekton/src/tasks/gitops/gitops-deprovision-app-config.yml.j2
@@ -29,6 +29,9 @@ spec:
       type: string
     - name: mas_app_id
       type: string
+    - name: cluster_url
+      type: string
+      default: ""
     - name: argocd_url
       type: string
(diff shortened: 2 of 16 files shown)


racree (approved):

tekton/src/tasks/gitops/gitops-delete-kafka-config.yml.j2
```diff
@@ -55,10 +58,18 @@ spec:
         value: $(params.github_pat)
       - name: SM_AWS_REGION
         value: $(params.avp_aws_secret_region)
+      - name: OCP_SERVER
+        value: $(params.cluster_url)
```
andrercm, line 62: why we're using `$(params.cluster_url)` to set `OCP_SERVER`?
should this be setting `CLUSTER_URL` or the other way around?

tekton/src/tasks/gitops/gitops-deprovision-app-config.yml.j2
```diff
@@ -60,10 +63,18 @@ spec:
         value: $(params.force_removal)
       - name: MAS_APP_ID
         value: $(params.mas_app_id)
+      - name: OCP_SERVER
```
andrercm, line 66: same as above...
why we're using `$(params.cluster_url)` to set `OCP_SERVER`?
should this be setting `CLUSTER_URL` or the other way around?

tekton/src/tasks/gitops/gitops-deprovision-app-install.yml.j2
```diff
@@ -56,10 +59,18 @@ spec:
         value: $(params.force_removal)
       - name: MAS_APP_ID
         value: $(params.mas_app_id)
+      - name: OCP_SERVER
```
andrercm, line 62: same as above:
why we're using `$(params.cluster_url)` to set `OCP_SERVER`?
should this be setting `CLUSTER_URL` or the other way around?

andrercm (changes requested):
Just a quick question..

tekton/src/tasks/gitops/gitops-delete-kafka-config.yml.j2
padmankosalaram: value set in param cluster_url is set to env variable OCP_SERVER

andrercm (approved):
Regarding my questions, Padmanabhan said ok we'll be reusing the value already set from cluster_url to also set OCP_SERVER.
Approving this PR.

tekton/src/tasks/gitops/gitops-deprovision-app-config.yml.j2
padmankosalaram: same as above. The value set in param cluster_url is set to env variable OCP_SERVER

PR #830, opened 2024-03-08
[patch] move cv-scan to end of the fvt pipelines to bring all jobs under scanner

sekharcvalluri:
At the moment cv scan is running before fvt in all pipelines and found that sometimes the cvscan runs before some jobs are created, so we end up with runs where some of the scans include all the jobs, and some don't. This PR is created to merge changes where cv-scan is moved to the end of the pipeline (finally block in all fvt pipelines)


rawa-resul:
FYI this is a continuation of https://github.com/ibm-mas/cli/pull/818. Discussion here: https://ibm-watson-iot.slack.com/archives/C04T6HQCHCP/p1709813508495699

rawa-resul (approved):

terenceq (approved):

Issue #599, opened 2023-11-07
CLI should be able to create empty config dir


durera:
When providing the local MAS config dir, the cli will currently exit if the provided directory doesn't exist, it should instead automatically create the directory (probably with a confirmation), e.g.

```
Directory /tmp/masconfig does not exist, create directory? Y/n
```

terenceq:
Closing as the PR:

https://github.com/ibm-mas/cli/pull/835

will address the issue.

PR #835, opened 2024-03-12
[patch] Create $LOCAL_DIR if the provided one does not exist

RafaAlFr:
When providing the local MAS config dir, the cli will currently exit if the provided directory doesn't exist, it should instead automatically create the directory

issue: [599](https://github.com/ibm-mas/cli/issues/599)

image
image
image


andrercm (approved):
LGTM

PR #841, opened 2024-03-14
[patch] add missing finally section to uninstall pipeline

creyesibm:
"Finally" section in the uninstall pipeline was not defined therefore it was not gathering logs and the pipeline seems to never finalize on the dashboard even when all task completed.
executed cli/tekton/test.sh and all tasks were added successfully, in this case here is the proof for the task modified under this PR.
image
March 18 uninstall pipeline containing this change ran successfully https://dashboard.masdev.wiotp.sl.hursley.ibm.com/tests/fvtuninstall


alequint (changes requested):
@creyesibm pls add in PR's description the test evidence of this new finally section. We need, to be sure, before anything, that the tekton resources can be applied without errors after your change doing what we discussed about `cli/tekton/test.sh`. Once they are good, we test that in uninstall pipeline before merging

alequint (approved):

alequint:
Looks great! :+1: 

image

Merging...

PR #855, opened 2024-03-21
[minor]  Capture DB2 and kafka versions

creyesibm:
Added code to get db2 cluster version and db2 operator version, also added code to find the kafka provider and its version. 
I will add test evidence once we run the code in the pipeline, local tests work as expected.


creyesibm:
@alequint @durera the code is ready, there seems to be a build issue right now but it has nothing to do with the code itself, it is a problem trying to upload the docker image. If you could please review so that I can work on the changes to have them ready for when the build works. Thank you

alequint:
@creyesibm let's redirect to he PR to managephase5 and we can test that in fvtnocpd. Then you can check if your version of dashboard will visualize the versions as expected. If that work we can go ahead and merge dashboard changes :+1:

alequint:
@creyesibm also, please rename your branch to be all lower case, I think that is what is causing the erorrs in CLI build

alequint:
Replaced by https://github.com/ibm-mas/cli/pull/858

PR #870, opened 2024-04-03
[patch][mascore-2107] Add support for MAS_SEGMENT_KEY

racree:
Add support for BasCfg segmentKey

### DIFF
--- a/image/cli/mascli/functions/gitops_mas_config
+++ b/image/cli/mascli/functions/gitops_mas_config
@@ -39,6 +39,7 @@ DRO Configuration (required iff MAS_CONFIG_TYPE is "bas"):
   --dro-contact-firstname ${COLOR_YELLOW}DRO_CONTACT_FIRSTNAME${TEXT_RESET}     The first name to register with DRO
   --dro-contact-lastname ${COLOR_YELLOW}DRO_CONTACT_LASTNAME${TEXT_RESET}       The last name to register with DRO
   --dro-ca-certificate-file ${COLOR_YELLOW}DRO_CA_CERTIFICATE_FILE${TEXT_RESET} The location of a file containing the DRO CA certificate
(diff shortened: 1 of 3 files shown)


image/cli/mascli/functions/gitops_mas_config
```diff
@@ -647,9 +651,16 @@ function gitops_mas_config() {
...
-      export SECRET_KEY_DRO_URL=${ACCOUNT_ID}${SECRETS_KEY_SEPERATOR}${CLUSTER_ID}${SECRETS_KEY_SEPERATOR}dro#dro_url
+      export SECRET_NAME_DRO=${ACCOUNT_ID}${SECRETS_KEY_SEPERATOR}${CLUSTER_ID}${SECRETS_KEY_SEPERATOR}dro
+      export SECRET_KEY_DRO_API_TOKEN=${SECRET_NAME_DRO}#dro_api_token
+      export SECRET_KEY_DRO_URL=${SECRET_NAME_DRO}#dro_url
+      export SECRET_KEY_MAS_SEGMENT_KEY=${SECRET_NAME_DRO}#mas_segment_key
       export DRO_CA_CERTIFICATE=$(cat ${DRO_CA_CERTIFICATE_FILE})
+      
+      if [[ -n "${MAS_SEGMENT_KEY}" ]]; then
+        sm_login
+        sm_update_secret $SECRET_NAME_DRO "{\"mas_segment_key\": \"$MAS_SEGMENT_KEY\"}"
```
whitfiea: If we update the secret named `SECRET_NAME_DRO` then it will override anything in that secret with **"{\"mas_segment_key\": \"$MAS_SEGMENT_KEY\"}"** this means we might potentially lose the dro_url and dro_api_token that might be set. We would need a distinct secret for this rather than a key in an secret

whitfiea (changes requested):

whitfiea (approved):

PR #872, opened 2024-04-04
[minor] Capture cp4d and components versions

creyesibm:
added code to capture CP4D version and versions from its components.
Tested the changes from the dashboard and everything looks good:
image

For environments without CP4D then a simple dash will be displayed without a pop-up.
image


alequint (approved):

alequint:
This is awesome @creyesibm ! And it looks good to me!

We need to keep an eye on it when we start using next CPD @andrercm had been working on... because versions may not be obtained from the same place then. But I see the code here is a tolerance to versions not found, so we should be safe for now.

alequint (approved):

Issue #867, opened 2024-04-01
Non-interactive uninstall of MAS instance is asking prompts for confirmation on cluster and instance id


NatarajBTI:
We are trying to achieve uninstall of MAS instance in non-interactive way using Helm chart and this is a requirement in ongoing UnifyBlue project.
But we are facing errors while trying it as it's throwing this error --> "Error: The project "mas--core"(B[m does not exist.:

Upon further checking, we see that instance id isn't picked up because it's expecting input confirmation on the following two prompts:
Proceed with this cluster? [Y/n] Y
MAS Instance ID > natinst1

Providing confirmation inputs to above two prompts isn't possible from code.

We have tried the following commands:

docker run -ti --rm -v ~:/mnt/home --pull always quay.io/ibmmas/cli

mas uninstall -i natinst1 --no-confirm

We have tried setting below values too:

export IBM_CATALOGS_ACTION="uninstall"
export COMMON_SERVICES_ACTION="uninstall"
export CERT_MANAGER_ACTION="uninstall"
export CLUSTER_MONITORING_ACTION="uninstall"
export UDS_ACTION="uninstall"
export MONGODB_ACTION="uninstall"
export SLS_ACTION="uninstall"

![image](https://github.com/ibm-mas/cli/assets/57617912/42822383-634b-4054-8179-f7cc9c067e35)

![image](https://github.com/ibm-mas/cli/assets/57617912/01c1dfd9-83c1-471b-b8ad-6b736dfdf9c7)

![image](https://github.com/ibm-mas/cli/assets/57617912/b6f61651-8dd7-4f96-8e48-5f7b72ee74a2)

[mas-deploy-sync-role-8g6h5-mas-deploy.log](https://github.com/ibm-mas/cli/files/14825072/mas-deploy-sync-role-8g6h5-mas-deploy.log)

whitfiea:
We should check here https://github.com/ibm-mas/cli/blob/ecc72f6c23a8b2f30b9e366500daa7ec08cd9ad0/image/cli/mascli/functions/uninstall#L24 is the MAS_INSTANCE_ID is set (which it should be in non-interactive mode) and just set the MAS_NS without prompting.

andrercm:
Should be fixed with the PR: https://github.com/ibm-mas/cli/pull/878

PR #878, opened 2024-04-05
[major] Fixes for mas uninstall command, remove `-id` option and change cert-manager removal

andrercm:
Available as preview in `cli:7.19.0-pre.mas-uninstall-fixes`

This PR improves the `mas uninstall` command:

- Introduces new arguments to support complete uninstall for MAS dependencies individually or all in a single run.
- Fix the issue while using `--no-confirm` that was prompting for cluster login and MAS instance ID. https://github.com/ibm-mas/cli/issues/867

Also see PR https://github.com/ibm-mas/ansible-devops/pull/1262 that changes the cert-manager removal process.

Usage:

1. `--uninstall-all-deps` argument includes all MAS dependencies to be uninstalled:

image

2. `--no-confirm` argument skips MAS instance id and cluster login confirmation:
image

Although if no target cluster identified, then it will still prompt for login:

image

image

** Uninstall pipeline ran fine:**

image


whitfiea (approved):

whitfiea:
Change to major as the `--id` option has been replaced with `--mas-instance-id` also we are now bringing in https://github.com/ibm-mas/ansible-devops/pull/1262 alters the delettion process for cert-manager

PR #880, opened 2024-04-08
[patch] certmanager health checkup

padmankosalaram:


image/cli/mascli/templates/gitops/bootstrap/argocd.yaml
```diff
@@ -1014,6 +1014,44 @@ spec:
...
+          hs = {}
+          if obj.status ~= nil then
+            if obj.status.conditions ~= nil then
+              for i, condition in ipairs(obj.status.conditions) do
+
+                if condition.type:endsWith('deploymentAvailable') then
+                  if condition.status == "True" then
+                    hs.status = "Healthy"
+                    hs.message = condition.message
+                    return hs
```
whitfiea: This will just find the first deploymentAvailable status that is True and return Healthy, even if the other deploymentAvailable are False.

whitfiea (changes requested):

whitfiea (approved):

PR #902, opened 2024-04-16
[minor] adding civil to mobile pipeline

gabrielbonamicoibm:
Screenshot 2024-04-16 at 15 41 10

### DIFF
--- a/tekton/src/tasks/fvt-launcher/launchfvt-mobile.yml.j2
+++ b/tekton/src/tasks/fvt-launcher/launchfvt-mobile.yml.j2
@@ -87,6 +87,13 @@ spec:
               name: mas-fvt-mobile
               key: FVT_DIGEST_MOBILE_TESTNG
               optional: false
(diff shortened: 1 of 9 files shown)


tekton/src/tasks/fvt-launcher/launchfvt-mobile.yml.j2
```diff
@@ -87,6 +87,13 @@ spec:
               name: mas-fvt-mobile
```
alequint, line 87: We have a Civil pipeline I created some time ago which is not used today... can we remove it as part of this PR of CLI, as now mobile will contain Civil things? (pipeline, launcher task, launcher playbook, etc?)

alequint (changes requested):

alequint (approved):

tekton/src/tasks/fvt-launcher/launchfvt-mobile.yml.j2
alequint: Civil pipeline will coexist for some runs but will be removed later

PR #906, opened 2024-04-18
Fix issues in finializer and cleanup the app config function for manage

whitfiea:
Fix issues in finializer and cleanup the app config function for manage

Rather than have every single option as a parameter for the function, we should accept the ws spec yaml file as it is passed to the function and apply that. Otherwise we even up having to update the function when we want to add a new entry to the cr.


image/cli/mascli/functions/gitops_suite_app_config
```diff
@@ -16,13 +16,14 @@ GitOps Configuration:
   -c, --cluster-id ${COLOR_YELLOW}CLUSTER_ID${TEXT_RESET}                                                 Cluster ID
       --gitops-config-dir ${COLOR_YELLOW}GITOPS_CONFIG_DIR${TEXT_RESET}                                   Directory that holds configuration to use
 
-
-IBM CP4D Service:
-      --cpd_product-version ${COLOR_YELLOW}CPD_PRODUCT_VERSION${TEXT_RESET}                             CP4D Product Version
-      --cpd-instance-namespace ${COLOR_YELLOW}CPD_INSTANCE_NAMESPACE${TEXT_RESET}                       CP4D Instance namespace to use
+IBM CP4D Service/Predict:
+  --cpd_product-version ${COLOR_YELLOW}CPD_PRODUCT_VERSION${TEXT_RESET}                                  CP4D Product Version
```
andrercm: typo here? `--cpd_product-version` should be `--cpd-product-version` ?

andrercm (changes requested):

image/cli/mascli/functions/gitops_suite_app_config
whitfiea: thanks, just fixed it

andrercm (approved):

PR #911, opened 2024-04-19
[patch] make DRO install default on newer versions of cli and mas catalog

lokesh-sreedhara:
- make DRO install default on newer version of cli and mas catalog
- enhance the DRO migration condition to support default catalog versions


andrercm (approved):
lgmt

PR #914, opened 2024-04-23
[patch] lookup mongo version not default during mas update

whitfiea:
The mas update command was looking up the mongo_extras_version_default in the casebundle which is always set to v5 for the current catalogs. What this should be doing is look up the mongo_extras_version which is the value of MONGODB_VERSION (or if not set then the default).

Without this fix the error message during mas update when you have a v6 mongodb is: 
```
MongoDB Community Edition is currently running on version 6.0.10 and cannot be downgraded to target version 5.0.23, which is the MongoDB version you would get when choosing MAS catalog 'v8-240405-amd64'
```


sanjayprab (approved):
good spot.

PR #918, opened 2024-04-24
[minor] [MASCORE-2307] support for execution of suite_dns role with c…

padmankosalaram:
Issue - https://jsw.ibm.com/browse/MASCORE-2307

What got changed in this PR:
 - Pipeline gitops-mas-initiator-after-provision invokes Task gitops-mas-initiator. In this task, added logic to create CIS AWS Secret which will hold the IBM Cloud API key
- Workaround fix , pause 2 min was added b/w 2 ibmcloud cis calls in `image/cli/install-ansible/ibm-mas_devops.tar.gz`
- Updated Task `gitops-suite `,  Pipeline `gitops-provision-mas-instance` with CIS specific params


whitfiea (changes requested):
@padmankosalaram Looking at the changes it seems that we don't have any default set for some of these new params on the pipeline and tasks. So things like cis_email might not be set as cis is not the dns_provider or no dns is expected. I think the result would be that "no default values set" if my saas-envs has no dns.cis_email. the same is true for others without a default.

whitfiea (approved):

PR #929, opened 2024-04-30
[patch] Fix verify-cli-pr git action to be trigger on push

andrercm:
Changing the new recent `verify-cli-pr` github action to be triggered on pushes instead of PR creations.

Before the fix it was failing during the code push:

image

After the fix - we skip push verification on dev branches:

image

### DIFF
--- a/.github/workflows/verify-cli-pr.yml
+++ b/.github/workflows/verify-cli-pr.yml
@@ -1,8 +1,9 @@
 name: Verify CLI Image
 
 on:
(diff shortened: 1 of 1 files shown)


whitfiea (commented):
@andrercm I think the pull_request on master is valid. In this situation you have a draft PR to master open so it is correct saying that it has an issue as there is a ibm-devops.tar.gz. Ia m not sure changing it to a push will work for a PR.

PR #933, opened 2024-05-02
[patch] Enable imagescan support in all applications

darshan-kansara1:
Epic : https://jsw.ibm.com/browse/MASCORE-881
Story : https://jsw.ibm.com/browse/MASCORE-954

Description
Adding a support of imagescan in below applications.

- Assist
- IOT
- Manage
- Monitor
- Optimizer
- Predict
- Visual inspection 

Related Issues
Depends on : https://jsw.ibm.com/browse/MASCORE-953
Depended on by : https://jsw.ibm.com/browse/MASCORE-956


tekton/src/pipelines/fvt-assist.yml.j2
```diff
@@ -78,6 +78,18 @@ spec:
         - input: "$(params.mas_app_channel_assist)"
           operator: notin
           values: [""]
+    - name: imagescan
```
durera: the name needs to be seet to `imagescan-{appId}`, e.g. imagescan-assist in this case .. same applies to all the other files as well

tekton/src/pipelines/fvt-assist.yml.j2
```diff
@@ -78,6 +78,18 @@ spec:
...
+      {{ lookup('template', 'taskdefs/ivt-core/common/taskref.yml.j2') | indent(6) }}
+      params:
+        {{ lookup('template', 'taskdefs/ivt-core/common/params.yml.j2') | indent(8) }}
+        - name: product_id
+          value: ibm-mas
```
durera: product_id needs to be updated too ... you can get the product_id value from the entry for ivtcore-xxx

Also need to add entry for product_channel same as we have for ivtcore too  (and check it's not missing for the core one while we're doing this) ... 

```
        - name: product_channel
          value: $(params.mas_app_channel_assist)
```

tekton/src/pipelines/fvt-data-dictionary.yml.j2
```diff
@@ -26,3 +26,15 @@ spec:
     # 1. ADD FVT
     # -------------------------------------------------------------------------
     {{ lookup('template', 'taskdefs/fvt-data-dictionary/data-dictionary.yml.j2') | indent(4) }}
+    - name: imagescan
```
durera: I don't think we can add this for data dictionary yet .. it's not a real MAS app, would leave it out for now ... get the apps done first.  this will need special handling

tekton/src/pipelines/fvt-sls.yml.j2
```diff
@@ -32,3 +32,15 @@ spec:
     # 1. SLS FVT
     # -------------------------------------------------------------------------
     {{ lookup('template', 'taskdefs/fvt-sls/sls.yml.j2') | indent(4) }}
+    - name: imagescan
```
durera: I don't think we can add this for sls yet .. it's not a real MAS app, would leave it out for now ... get the apps done first.  this will need special handling

durera (changes requested):

durera (approved):

PR #934, opened 2024-05-02
Remove redundant/unused values from gitops templates. Update rootapp template to reflect param changes.

tomklapiscak:


tomklapiscak:
reraised as https://github.com/ibm-mas/cli/pull/940 (the branch for this PR has an upper case char in its name, which isn't valid apparently)

PR #947, opened 2024-05-09
[patch] Add mongodb_version to update pipelinerun

whitfiea:
The mongodb_version was missing from the pipelinerun for update, so when the cli sets it here https://github.com/ibm-mas/cli/blob/c602b9db623448dc8dd324da76c4ae3433db4032/image/cli/mascli/functions/update#L686 it is then not set in the pipelinerun for update as it was missing.

### DIFF
--- a/image/cli/mascli/templates/pipelinerun-update.yaml
+++ b/image/cli/mascli/templates/pipelinerun-update.yaml
@@ -27,6 +27,8 @@ spec:
       value: '$MONGODB_NAMESPACE'
     - name: mongodb_v5_upgrade
       value: '$MONGODB_V5_UPGRADE'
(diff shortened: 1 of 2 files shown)


HarveyTangg:
I found that 'mongodb_version' is also missing in the 'mas-update' pipeline definition. So, probably /tekton/src/pipelines/update.yml.js also needs to be updated?

andrercm (dismissed):
good catch

andrercm:
can you also add `MONGODB_V6_UPGRADE` in there?

whitfiea:
@andrercm can you review again please?. This is now verified to work.

andrercm (approved):

PR #950, opened 2024-05-10
[minor] support option to --skip-grafana-install

sanjayprab:
This PR adds the argument `--skip-grafana-install` that allows skipping grafana  but avoiding to check if Grafana is available in the cluster based on the available of the operator package. and sets `CLUSTER_MONITORING_INCLUDE_GRAFANA=False`

This has been implemented in:

mas install

### DIFF
--- a/image/cli/mascli/functions/install
+++ b/image/cli/mascli/functions/install
@@ -753,8 +753,9 @@ function install() {
     esac
   done
 
-  if [ $SKIP_GRAFANA_INSTALL == "true" ]; then
+  if [[ "$SKIP_GRAFANA_INSTALL" == "true" ]]; then
     CLUSTER_MONITORING_INCLUDE_GRAFANA="False"
+    echo "Skipping Grafana..."
   else
     # Determine whether Grafana is available in the cluster based on the available of the operator package
     oc get packagemanifest grafana-operator &> /dev/null
@@ -764,6 +765,8 @@ function install() {
       CLUSTER_MONITORING_INCLUDE_GRAFANA="True"
     fi
   fi
+
+  echo "Grafana env var(CLUSTER_MONITORING_INCLUDE_GRAFANA) = ${CLUSTER_MONITORING_INCLUDE_GRAFANA}"
 
   if [[ "$INTERACTIVE_MODE" == "true" ]]; then
     load_config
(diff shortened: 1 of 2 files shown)


image/cli/mascli/functions/install
```diff
@@ -744,18 +744,25 @@ function install() {
       --dev-mode)
         DEV_MODE=true
         ;;
+      --skip-grafana-install)
+        SKIP_GRAFANA_INSTALL=true
```
andrercm: what if we do:
```
--skip-grafana-install)
CLUSTER_MONITORING_INCLUDE_GRAFANA=false
```

image/cli/mascli/functions/install
```diff
@@ -744,18 +744,25 @@ function install() {
...
       -h|--help)
         install_help
         ;;
     esac
   done
 
-  # Determine whether Grafana is available in the cluster based on the available of the operator package
-  oc get packagemanifest grafana-operator &> /dev/null
-  if [[ "$?" == "1" ]]; then
+  if [[ "$SKIP_GRAFANA_INSTALL" == "true" ]]; then
```
andrercm: and then here we just do a check to see if `CLUSTER_MONITORING_INCLUDE_GRAFANA` is defined already, if not defined then we set it upon packagemanigest logic
```
# if CLUSTER_MONITORING_INCLUDE_GRAFANA is not defined 
if [[ -z "$CLUSTER_MONITORING_INCLUDE_GRAFANA" ]]; then
  oc get packagemanifest grafana-operator &> /dev/null
  if [[ "$?" == "1" ]]; then
    CLUSTER_MONITORING_INCLUDE_GRAFANA="False"
  else
    CLUSTER_MONITORING_INCLUDE_GRAFANA="True"
  fi
fi
```

andrercm (changes requested):
added some thoughts to it

image/cli/mascli/functions/install
sanjayprab: used it

durera (changes requested):
I have overhauled cluster_monitoring / grafana in the pending major update .. can you take a look at how to integrate this with that branch (the cluster_monitoring stage isn't even in the pipeline anymore, there's a dedicated grafana stage).

I think the action is to work with the GRAFANA_ACTION env var in the update, and ensure it can be set explicitly instead of only being able to be set automatically.

sanjayprab:
@durera closing this PR and have created a new PR  with a new branch  due to various conflicts.

PR #954, opened 2024-05-13
[minor] support option to --skip-grafana-install in mas install function

sanjayprab:
This PR adds the option `--skip-grafana-install` that allows skipping grafana by avoiding to check if Grafana is available in the cluster based on the available of the operator package. and sets GRAFANA_ACTION="none"

This has been implemented in:

mas install


durera (approved):
Looks good, will merge it shortly as discussed in Slack

PR #958, opened 2024-05-14
Use shorter name for ArgoCD Apps

whitfiea:
Follows https://github.com/ibm-mas/gitops/pull/79 to uise shorter names which is mainly used in the app checking in FVT and also the lock message

https://jsw.ibm.com/browse/MASCORE-2682


image/cli/mascli/functions/gitops_utils
```diff
@@ -281,12 +281,10 @@ function unlock_git_repo() {
 function git_lock_branch_name() {
 
   LOCK_NAME=$1
-  ACCOUNT_ID=$2
```
tomklapiscak, line 284: I'm not sure it's safe to remove the ACCOUNT_ID from the lockbranch name; the lock effectively being acquired by this is global to the config repo (i.e. gitops-envs) and that can still contain config for >1 account.id. 

Removing the region.id is safe since we now rely on cluster ids being unique across all regions of an account. Although we're not actually enforcing this anywhere. As an aside, I wonder if we want to remove the region path component from our gitops-envs structure as this would effectively enforce unique cluster ids per region?

E.g.
```
mcsp-dev
   us-east-1
      cluster-1
   us-east-2
      cluster-1
```
would cause issues (as cluster id is not unique across all regions. If we removed the region path component, we'd force* a unique cluster id like so:

```
mcsp-dev
    cluster-1-us-east
    cluster-1-us-east-2
```

(*uniqueness of cluster id isn't actually enforced, since the ID and path component name can be different at present, but it at least emphasises the point)

whitfiea: Thasnks, I added the ACCOUNT_ID back into the lock so we can protect that.

For the region_id I have removed the region from the path and the same update in the gitops repo

image/cli/mascli/functions/gitops_utils
```diff
@@ -282,11 +282,10 @@ function git_lock_branch_name() {
...
   LOCK_NAME=$1
   ACCOUNT_ID=$2
-  REGION_ID=$3
-  CLUSTER_ID=$4
-  MAS_INSTANCE_ID=$5
+  CLUSTER_ID=$3
+  MAS_INSTANCE_ID=$4
 
-  echo -n "lock.${LOCK_NAME}.${ACCOUNT_ID}.${REGION_ID}.${CLUSTER_ID}"
+  echo -n "lock.${ACCOUNT_ID}.${LOCK_NAME}.${CLUSTER_ID}"
```
tomklapiscak: Minor point as I think this would work as-is, but I think I'd prefer to this be:
`"lock.${LOCK_NAME}.${ACCOUNT_ID}.${CLUSTER_ID}"`
to align with how we had it before (minus the region id).

whitfiea: changed

PR #965, opened 2024-05-16
[patch] adding suite-cert task to install pipeline

leo-miran:
Adding suite cert-task in install pipeline that is required when using manual certificates configuration
image


alequint (commented):
Good point, suite-certs was one of the gaps we had between official install pipeline used in CLI and the fvt pipeline. Another thing we will need to do besides adding that in the pipeline itself is interface that setting in CLI flow to assist use of manual certificates in the installation flow ... but it is something to be designed and thought more. For now, having this task there will not be a problem it will just be skipped in regular runs of CLI... but it is needed in our fvts if we want to use manual certs... changes look good to me PR

alequint (approved):

Issue #856, opened 2024-03-22
Lack of documentation for mirroring Maximo IT images (mirror-images)


SalaryTheft:
I think `--mirror-icd` option should be explained in the `mirror-images` documentation.
(It is the only industry solution(add-on) that can be selectively mirrored, I assume.)
https://github.com/ibm-mas/cli/blob/master/docs/commands/mirror-images.md

PR #973, opened 2024-05-18
[patch] Add missing Maximo IT command flag to doc

andrercm:
Fix for : https://github.com/ibm-mas/cli/issues/856


whitfiea (dismissed):
Does --mirror-icd actually do anything? I can't see where it is used other than in this doc page?

andrercm:
@whitfiea I haven't implemented it, just fixing the docs for it, but it should be used by the `mirror_add_manage` playbook to mirror Maximo IT images along with Manage's images:

https://github.com/ibm-mas/ansible-devops/blob/master/ibm/mas_devops/playbooks/mirror_add_manage.yml#L34

durera:
> Does --mirror-icd actually do anything? I can't see where it is used other than in this doc page?

Yup, it does ... it controls the setting of the `MIRROR_MAS_ICD` env var, which the `mas mirror-images` command passes through to the playbook that drives the image mirroring process.

PR #984, opened 2024-05-22
[patch] change log file variable to match name in utils

mattlrx:
change log file variable to match name in utils 
the must gather was giving error like:
[4mCollect Reconcile Logs(B[m
/mascli/must-gather/../functions/internal/utils: line 60: $LOGFILE: ambiguous redirect


durera:
These are different log files .. this is already fixed in https://github.com/ibm-mas/cli/pull/977

- LOGFILE is the log for the CLI generally, and is additive
- LOG_FILE is the log for one specific run of must-gather, which is added to the must-gather itself

PR #990, opened 2024-05-24
[minor] [mascore-2637] support for manual_cert_mgmt: "True"

padmankosalaram:
Issue - https://jsw.ibm.com/browse/MASCORE-2637?filter=-1

1. Manual certs will be added to saas-envs (saas-envs/gitops/aws-dev/us-east-2/fvtsaas/fvtsaas2/manual-certs)
2. The multiple Crt/key files will be combined into an YAML, a bash script does this. (saas-envs/create-aggregate-files.sh)

3. Further the YAML file will be consumed by 01-ibm-mas_suite_certs_Job (gitops/instance-applications/010-ibm-sync-jobs/templates/01-ibm-mas_suite_certs_Job.yaml) which creates crt/key from the yaml file 
and invokes suite_certs role.
4. Also core/app specific secret having the tls.crt / tls.key ca.crt will be created in gitops


image/cli/mascli/functions/gitops_suite
```diff
@@ -565,6 +558,26 @@ function gitops_suite() {
   fi
   mkdir -p ${GITOPS_INSTANCE_DIR}
 
+  # Set certs only if manual cert is True (to create k8s secret in gitops)
+  # ---------------------------------------------------------------------------
+  
+  if [ "${MAS_MANUAL_CERT_MGMT}" == "True" ] ; then
+    if [[ -n "$MAS_MANUAL_CERTS_YAML" && -s "$MAS_MANUAL_CERTS_YAML" ]]; then
```
whitfiea, line 565: what does the -s option do here? Should that be -f to make sure the file exists?

whitfiea (changes requested):

image/cli/mascli/functions/gitops_suite
padmankosalaram: Added -s (instead of -f) as the logic checks for a file with nonzero size

-s filename - Check if file is nonzero size
-f filename - Check for regular file existence not a directory

whitfiea (approved):

PR #995, opened 2024-05-27
[patch][MSO-1505] Update fvt-optimizer pipeline to remove optimizer-manage-user step

lgustavoibm:
Update Pipeline test from FVT Env to remove the creation of user in Selenium (fvt-optimizer-manage-user), the team of FVT has created a new script to to this step using pytest, so it's only necessary to run using that credentials.


andrefgj (dismissed):

andrefgj (dismissed):
It looks good to me

andrercm (approved):
looks good

PR #997, opened 2024-05-28
[patch] Install COS and Kafka when Manage is present

alequint:
... and when COS and Kafka appropriated parameters are informed.

This is needed because Manage can optionally use Kafka and COS internally. And we have FVTs to test these cases.

----

Tested in pfvtcisquad, run 1167: https://dashboard.masdev.wiotp.sl.hursley.ibm.com/tests/pfvtcisquad


alequint:
dup of https://github.com/ibm-mas/cli/pull/1010

PR #1004, opened 2024-05-29
[minor] Support Maximo Location Services for Esri

andrercm:
This PR orchestrates the install and pipeline enable of new addon named Maximo Location Services (under the hood it deploys Esri ArcGIS Enterprise) which should be installed only when Manage and Maximo Spatial is installed.

image

image


alequint (commented):
All the three PRs look good to me @andrercm 

I have just some comments/questions though:

- "**Offline**": I see you had updated CLI flow to include arcgis optional component when Spatial is installed... before we release this, shouldn't we have arcgis curated in our latest catalog in case someone wants to install that in an off line cluster? guarantee we can mirror and also test that in a fvtairgap? How does that work in cases like this @durera @sanjayprab ?

- "**RunAfter**": I see you placed arcgis task after manage-install... not a big deal, but usually we do not install dependencies in the middle of the application  (between install and config)... we can put that task behind a bit, if not in the beginning like nvidia  (when not so disruptive like cpd) or after suite-verify...

- **Configuration**: more important than the order when tasks runs: should arcgis be up and running when Spatial is installed and/or started?  How is Spatial configured against arcgis? I presume we will have no Cfg CR and we instead some automation similar to the one to configure attachments? Is that the idea?

andrercm:
**Offline**: It was agreed that there's no support for airgap at this 9.0 release for MLS, this will be included in plan for future 9.1
**RunAfter**: That's a good point and my thinking on adding that only after Manage operator install was to ensure that MLS is really installed in parallel with Manage activation... but as there's no direct dependency between then, we could also add MLS to be installed right after MAS core is installed (that's the only real dependency for MLS)... what you think?
**Configuration**: That's right, at this 1st release, MLS and MAS are not directly attached in terms of configuration, although it requires MAS Core to be installed. The arcgis step will only ensure that the operator is in place and that the arcgis enterprise server is up. At this 1st moment, it will require some additional manual steps to fully configure it, also items that should be addressed and improved within next MLS releases.

alequint (dismissed):
Looks good to me

whitfiea (approved):

PR #1007, opened 2024-05-31
[minor] [MASCORE-2438] Allow db2 database storage size to be set

twhart:
Added a couple changes to allow storage size to be set for db2u 

image


tekton/src/pipelines/gitops/gitops-mas-apps.yml.j2
```diff
@@ -61,6 +61,20 @@ spec:
...
+    - name: db2_temp_storage_size
+      type: string
+    - name: db2_logs_storage_size
+      type: string
+    - name: db2_audit_logs_storage_size
+      type: string
+    - name: db2_data_storage_size
+      type: string
+    - name: db2_backup_storage_size
+      type: string
```
whitfiea, line 81: We should set the default on these to be "" so that we don't have to set them and just use the default if not present.

whitfiea (changes requested):
Can you remove the ibm-mas_devops.tar.gz file?

tekton/src/tasks/gitops/gitops-db2u-database.yml.j2
```diff
@@ -36,6 +36,18 @@ spec:
...
+    - name: db2_temp_storage_size
+      type: string
+    - name: db2_logs_storage_size
+      type: string
+    - name: db2_audit_logs_storage_size
+      type: string
+    - name: db2_data_storage_size
+      type: string
+    - name: db2_backup_storage_size
+      type: string
```
twhart, line 50: @whitfiea should we add defaults on to these as well?

whitfiea: those should inherit the default from the pipeline, so it isn't needed unless the task is used in a different pipeline

twhart: Ah okay got it, this should be all set then

whitfiea:
@twhart how about the "Can you remove the ibm-mas_devops.tar.gz file?" https://github.com/ibm-mas/cli/pull/1007#pullrequestreview-2093295047

twhart:
> @twhart how about the "Can you remove the ibm-mas_devops.tar.gz file?" [#1007 (review)](https://github.com/ibm-mas/cli/pull/1007#pullrequestreview-2093295047)

Should be set now, sorry I saw that and realized after

whitfiea (approved):

PR #1009, opened 2024-06-03
[minor] Add install sync point at the end of fvt-launcher #996

unnati-solanki-git:
Refer to https://jsw.ibm.com/browse/MASCORE-2830

In summary, even though predict had been successfully configured, its deployments, afterwards, start failing and fvt launcher was not able to figure out; it's because we assume installation works and we have to stop fvt at the end of that pipeline. A consequence is that all tasks passed, fvt launcher is finalized and cluster is then de-provisioned.

A waitfor task has been added after all fvt approval tasks and before finally step
This task waits until sync-install has INSTALL_STATUS defined. Failures are not ignored and no escape routes defined
"En Passant" fix

In addition, small fix in wait-for-cm task: before checking if escape condition is reached, we check if expected wait/for result is achieved; instead we should check if expected result was not achieved (like the other escape validation we have in the same code about target value)

✅ Tested in fvt811x (run 2922), fvtstable (run 2925) and fvtcpd (run 2919)


unnati-solanki-git:
This one is dup of https://github.com/ibm-mas/cli/pull/996 and 996 is merged now. Hence, closing this one. 
Thanks!

PR #1010, opened 2024-06-03
[patch] Install COS and Kafka when Manage is present

unnati-solanki-git:
... and when COS and Kafka appropriated parameters are informed.

This is needed because Manage can optionally use Kafka and COS internally. And we have FVTs to test these cases.

Tested in pfvtcisquad, run 1167: https://dashboard.masdev.wiotp.sl.hursley.ibm.com/tests/pfvtcisquad


alequint (approved):

alequint:
Some more context discussed in slack about this PR. This change would be a part 1 of 2 parts:

----

I was trying to think in a two parts push here as we need COS for Manage and Mobile FVTs, considering COS (and Kafka, which came as bonus here in the PR) is actually an optional service to be used:

- Part 1 (current PR): would be just to make sure COS is part FVT environments, so correspondent FVTs can connect to those services and do what they need. This change would not impact existing CLI flows, because besides installing Manage, pipeline would need to setup cos_type or kafka_action params to make it work (and so pass the task `when` clauses), which we do not have have in manage installation flow in CLI today (it does in Assist or in IoT/Monitor side only, respectively).

- Part 2: then we would need to bring to CLI flow the optional use of COS or Kafka when someone installs Manage. That flow will set cos_type and kafka_action properly during Manage flow (and also the different type of configs). I just let this one to be done once 1 is in use for FVT because it consumes more time to develop and test - still in this part 2 would be required the enhancement of suite-app-config we talked elsewhere, where we discuss the multiple manage settings that happen after ManageWorkspace is installed (we would need the enhanced suite-app-config setup CR upfront, instead of setting it later causing unnecessary reconciles).

Giving the above, how you feel about these two parts being handled separately @durera ? (installation as part 1 with no impact in existing flow X configuration a part 2, to parameterize pipeline properly from Manage CLI installation flow to also configure attachments feature to use it)

alequint (approved):

alequint:
Just one more comment about the run of CLI using this version of pipeline:

```
docker run -ti --rm -v ~/Documents/projects/mas/masconfig:/mnt/home --pull always quay.io/ibmmas/cli:9.2.1-pre.managecosus
```

Note just Manage is installed (no Assist, no IoT):

image

And in Pipeline, neither costype nor kafka action has been set, skipping COS and Kafka installation:

image

image

Part 2 mentioned above will be included section 8.1 of CLI, Configure Maximo Manage, so users can decide how to setup attachments.

PR #1015, opened 2024-06-06
[patch] must-gather file path change for upgrade pipeline

leo-miran:
Using in pod storage for must-gather file


alequint (dismissed):

alequint:
Updated to extend the fix to other layer 4 pipelines

alequint (approved):

PR #1040, opened 2024-06-18
[minor] DRO should default to ibm-software-central namespace

lokesh-sreedhara:
"DRO operator v2.16.0 is published and defaults namespace to ibm-software-central. the role bindings are created for that namespace, as well as the legacy default namespaces so, should not impact upgrade, and for new deployments, you can use ibm-software-central namespace if you wish."

This PR is to change the DRO default namespace from redhat-marketplace to  ibm-software-central

Related ansible PR: https://github.com/ibm-mas/ansible-devops/pull/1342


yuvraj-vansure (approved):

lokesh-sreedhara:
Change in default namespace is not required at this junction. closing this PR out

PR #1041, opened 2024-06-19
[patch] Update run-suite task with health test parameters

alequint:
After this PR, pls review and merge ansible-fvt: https://github.ibm.com/maximoappsuite/ansible-fvt/pull/436

Tested in fvtcpd:

image

Attributes added in manage secret (values are there, just not shown in screenshot):

image


tekton/src/tasks/fvt/fvt-run-suite.yml.j2
```diff
@@ -197,6 +197,26 @@ spec:
       # Common Test Framework Specific information
       - name: CTF_IS_LOCAL
         value: $(params.ctf_is_local)
+      
+      # Health Test Specific data
+      - name: RHM_APIKEY
+        valueFrom:
+          secretKeyRef:
+            name: mas-fvt-manage
```
leo-miran, line 205: @alequint I see this task is pulling from mas-fvt-manage secret. I believe these secrets are only created when manage fvt launcher is set to true. In a health standalone scenario this may not exist. Is this same task going to be executed in this scenario?

Maybe it should pull these values from a secret that is created only when health is configured to be launhched.

leo-miran (approved):

PR #1060, opened 2024-06-25
[patch] Add delete for jdbc cfg for manage

caroazad:
Add task and to pipeline for jdbc cfg deletion for manage


whitfiea (changes requested):
Do we need the ansible-devops tar file? Is there a saas-tekton change that goes with this?

caroazad:
> Do we need the ansible-devops tar file? Is there a saas-tekton change that goes with this?

No I don't need that file. I don't know how that happened

whitfiea (approved):

PR #1073, opened 2024-06-28
[patch] Skip CPD upgrade if not initially installed in ibm-cpd namespace

andrercm:
This is to fix the scenario where customer has installed cpd in a custom namespace other than `ibm-cpd`, in this case it current logic was installing a new cpd instance in the defaulted namespaces.
After this fix, if cpd not detected in `ibm-cpd` then we won't upgrade/install cpd at all.
Also added an alternative option to skip cpd upgrade upfront by setting `--skip-cp4d-upgrade`

image


andrercm:
Waiting customer to test the beta version of cli with this change

andrercm:
given recent changes done to convert cli from shell to python, this PR does not make sense anymore.

PR #1104, opened 2024-07-08
[patch] Fail task when embedded command fails

caroazad:
This is a fix for the bug reported in [MASCORE-3129](https://jsw.ibm.com/browse/MASCORE-3129) which seems to occur when there is an error in the function. 
Add a trap to catch errors which displays the line number where the error occurred before exiting.


whitfiea (changes requested):
the changes for the issue https://jsw.ibm.com/browse/MASCORE-3129 need to cover any unexpected errors from anywhere else in al the gitops functions as we need to stop unexpected errors from going missing in the task runs

whitfiea (approved):

PR #1112, opened 2024-07-12
[minor] Add support manual certificate management

harsh42774:
### Description:
This PR adds the support to add manual certificates required by the recently added suite-certs role (Already being used in FVTs).

For Interactive CLI Install, prompts are added as below:
If User responds "y" for "Configure domain & certificate management? [y/n]" prompt, they will be asked with the prompt for manual certificates as seen in the screenshot below:
Screenshot 2024-07-12 at 12 45 47

For Non-Interactive CLI Install, a new argument "manual-certificates" is added, which takes the value as the path containing the certificates.
Eg: --manual-certificates '/mnt/home/certs'

The above interactions will set the env variable "mas_manual_cert_mgmt" to True which is used by the suite-certs role. The Local directory value provided by the user will be checked if it exists and if the 3 relevant certificates are present for core and the apps selected for installation. If any file is missing, a fatal error will be triggered saying the name of the file missing and in which directory.
Upon successful validation of the certificates, secrets will be created using the change in python-devops (https://github.com/ibm-mas/python-devops/pull/10) (as they are currently created for FVTs).

### Related Links:
https://jsw.ibm.com/browse/MASCORE-2795
https://jsw.ibm.com/browse/MASCORE-3152
https://jsw.ibm.com/browse/MASCORE-3153

### Testing:
Tested in a quickburn cluster.

Secrets created for Core+Manage install:
Screenshot 2024-07-11 at 15 27 05

Suite-certs role succeeds after picking the manual certificates from the secret mounted:
Screenshot 2024-07-11 at 15 27 23

Public Certificate is created in Core namespace using the Core secret mounted:
Screenshot 2024-07-11 at 15 27 31

Public Certificate is created in Manage namespace using the Manage secret mounted:
Screenshot 2024-07-11 at 15 28 30


python/src/mas/cli/install/argParser.py
```diff
@@ -143,6 +143,11 @@ def isValidFile(parser, arg) -> str:
     action="store_const",
     const="false"
 )
+masAdvancedArgGroup.add_argument(
+    "--manual-certificates",
+    required=False,
+    help=""
```
durera: Please add a help message here (and update the help message that is included in docs/commands/install.md)

harsh42774: message added in both files

durera (approved):

PR #1122, opened 2024-07-16
[minor] Add gitops changes to cli

whitfiea:
This PR adds all the functions required for the gitops way of deploying MAS. See https://github.com/ibm-mas/gitops for details on this method.

The changes in the mas-cli are to add functions that will use the templates required to commit changes to a git repo that the helm charts in https://github.com/ibm-mas/gitops can use. The changes also include tekton pipelines and tasks to call those functions if required (the functions can be called without the need for the pipelines/tasks).

The tekton pipelines/tasks provide a layer of automation that can be used to call each function if required and to orchestrate the creation of git files and secrets.

The summary of changes in this PR are:
- remove the need for yq to be installed via pip
- Add cli and ansible-devops version to finializer as it is not added in gitops FVT run as we don’t call ansible roles, which is adding this data in the non-gitops FVT
- Add gitops related functions
- Update mustgather to get ArgoCD application and applicationset resources
- Add gitops related templates used by the gitops functions
- Removes python/src/mas/cli/templates/pod-templates/saas-essentials/ibm-mas-visualinspection.yml as this is not to be exposed
- Build tekton pipelines and tasks related to gitops
- Add pipeline_src_dir to all pipelines and tasks so that we can use this for the gitops sub-folder within tekton/src/pipelines and reference taskdefs
- Add gitops related tekton pipelines and tasks

**There is no change to the mas cli in terms of usage at this point in time. A user runing the cli will not see any options to select gitops method for example.**


durera (approved):
I don't agree with removing the pod pemplates from here as they provide a useful reference point to customers, but also can't be bothered to argue with the folks from the SRE org.

PR #1131, opened 2024-07-22
[patch] Remove MIF Tests From FVT Pipeline

chiragcsatani:
base-ui-mif-setup

base-ui-mif-kafka-setup

base-ui-mif-jms-inbound

base-ui-mif-kafka-inbound

Above tests are actively being converted to pytest. Hence removing from pipeline


alequint:
Your changes look good @chiragcsatani but we need to have some test evidence here to see how pipeline looks like after this change. What you can do for a quick review:

- oc login your internal fyre cluster
- in cli/tekton, run the two playbooks there:

```
ansible-playbook generate-tekton-pipelines.yml
ansible-playbook generate-tekton-tasks.yml
```

- in cli/tekton, run test.sh - it will connect in the cluster you logged in above, and will install all the resources created by playbooks above (they will be all created in cli/tekton/target) and install them in the default namespace of your cluster (which will not affect the pipelines, which run in thei own mas-xxx-pipelines namespace).

- Run it till the end and post evidence in the PR. Also, go to your cluster look for fvt-manage pipeline inside the default namespace to check if phases are preserved and if no error was introduced. Paste screenshot as evidence in this PR.

By doing this, creating the resources in the cluster without problem it means you will not have problems during setup_pipeline task in ansible-fvt. Otherwise it might block all the official pipelines to run

alequint (changes requested):
Test evidence requested: https://github.com/ibm-mas/cli/pull/1131#issuecomment-2243754022

unnati-solanki-git:
@chiragcsatani @alequint I attempted to run this locally as mentioned in the instruction and able to see the sequence of task/test like this: 

image

Let me know if you need anything else. Thanks!

chiragcsatani:
@alequint can you provide your feedback here @unnati-solanki-git 
if its okay then we can merge so jira item can be closed.

alequint (approved):

alequint:
Sequence of tasks look great... thanks for posting @unnati-solanki-git (and sorry, did not see it before, not following Git emails, relying on Slack notification to review the PRs 😄  )

PR #1145, opened 2024-07-27
[patch] Remove optimizer gs-resource-levelling test

alequint:
... as requested by Optimizer team, as this test does not work as it should and must be removed


andrercm (approved):
approved

PR #1149, opened 2024-07-30
[patch] Updates for manage attachments and logging for gitops

racree:
Updates the gitops functions to fix an issue with the manage and logging secrets used for gitops.

Also updates the gitops_cos function to configure AWS S3 via terraform which is only supported for internal IBM usage, and will not work outside of IBM as it uses internal modules.

The terraform cli is installed only when the internal IBM route is taken rather than install it in the base image.


image/cli/mascli/functions/gitops_deprovision_cos
```diff
@@ -185,19 +185,180 @@ function gitops_deprovision_cos() {
...
+
+    rc=$?
+    [ $rc -ne 0 ] && exit $rc
+
+    echo -e "Deleting secret  $COS_SECRET"
+    sm_delete_secret $COS_SECRET
+
+  elif [[ "$COS_TYPE" == "aws" ]]; then
+
+    curl https://releases.hashicorp.com/terraform/1.9.2/terraform_1.9.2_linux_amd64.zip --output $TEMP_DIR/terraform_1.9.2_linux_amd64.zip
```
whitfiea, line 204: Rather than install terraform here we should install it as part of the cli-base image here https://github.com/ibm-mas/cli-base/tree/stable/image/cli-base/install so that it is present all the time.

image/cli/mascli/functions/gitops_cos
```diff
@@ -161,34 +161,195 @@ function gitops_cos() {
...
+    ESCAPED_INFO=${ESCAPED_INFO//$'\n'/\\n}
+
+    sm_update_secret $COS_SECRET "{\"username\": \"$COS_USERNAME\", \"password\": \"$COS_PASSWORD\", \"info\": \"$ESCAPED_INFO\"}"
+    rc=$?
+    [ $rc -ne 0 ] && exit $rc
+
+
+  elif [[ "$COS_TYPE" == "aws" ]]; then
+
+    curl https://releases.hashicorp.com/terraform/1.9.2/terraform_1.9.2_linux_amd64.zip --output $TEMP_DIR/terraform_1.9.2_linux_amd64.zip
```
whitfiea, line 197: As per the other comment, move this install into the cli-base image

image/cli/mascli/functions/gitops_cos
```diff
@@ -18,12 +18,12 @@ Secrets Manager:
...
 
 IBM Maximo Application Suite:
-  -m, --mas-instance-id ${COLOR_YELLOW}MAS_INSTANCE_ID${TEXT_RESET}                                    IBM Suite Maximo Application Suite Instance ID  
+  -m, --mas-instance-id ${COLOR_YELLOW}MAS_INSTANCE_ID${TEXT_RESET}                                    IBM Suite Maximo Application Suite Instance ID
       --mas-config-dir ${COLOR_YELLOW}MAS_CONFIG_DIR${TEXT_RESET}                                      IBM Suite Maximo Application Suite config dir to hold config
 
 COS Provider:
       --cos-action ${COLOR_YELLOW}COS_ACTION${TEXT_RESET}                                                  provision or deprovision
-      --cos-type ${COLOR_YELLOW}COS_TYPE${TEXT_RESET}                                                      COS type, only ibm supported
+      --cos-type ${COLOR_YELLOW}COS_TYPE${TEXT_RESET}                                                      COS type, ibm or aws supported
```
whitfiea: The aws part is only supported internal to IBM, can we add that to this line as if an external customer tries `aws` then it will fail as it needed internal IBM terraform modules

whitfiea (changes requested):
Can we add a description to this PR?

image/cli/mascli/functions/gitops_deprovision_cos
racree: Terence stated it is something we shouldn't ship to customers.  Does that effect how it should be installed?  Would licensing effect the decision whether to put it in the cli-base or not?

whitfiea (dismissed):

whitfiea (approved):

PR #1168, opened 2024-08-09
[minor] Support ability to disable "WalkMe" guided tour feature

sekharcvalluri:
There is a system setting for "Guided Tours" (spec->settings->walkme) in Suite CR. It is enabled by default and can be configured from the admin UI. Due to the pop of "what's new" some of the UI test cases from Manage are failing, hence, that needs to be turned off for all fvt environments by default. As a fix, an environment variable (MAS_ENABLE_WALKME) has been introduced to control the 'walkme' setting. FVT pipelines need to use this env to disable the setting.

Testing: Changes have been tested locally as well as on the pfvt (fyre-envs) env.

Bug details and snap shots of the FVT run can be found here: MASCORE-3444

Reference to ansible-devops changes: https://github.com/ibm-mas/ansible-devops/pull/1409


durera (changes requested):
although the param is `mas_enable_walkme` I think at the CLI level we probably want the default if you do nothing with non-interactive to be walkme enabled ... so 

```
-mas-enable-walkme MAS_ENABLE_WALKME           Enable or disable guided tour
```

should probably be

```
-disable-walkme  Disable MAS guided tour
```

And the param would be configured as:

```
masAdvancedArgGroup.add_argument(
    "--mas-enable-walkme",
    dest="mas_enable_walkme",
    required=False,
    help="Enable or disable guided tour",
    action="store_const",
    const="false"
)
```

Much in the same way that `--disable-ca-trust` param works.  Other than that tweak to the command flags I think the rest is good 👍

sekharcvalluri:
> although the param is `mas_enable_walkme` I think at the CLI level we probably want the default if you do nothing with non-interactive to be walkme enabled ... so
> 
> ```
> -mas-enable-walkme MAS_ENABLE_WALKME           Enable or disable guided tour
> ```
> 
> should probably be
> 
> ```
> -disable-walkme  Disable MAS guided tour
> ```
> 
> And the param would be configured as:
> 
> ```
> masAdvancedArgGroup.add_argument(
>     "--mas-enable-walkme",
>     dest="mas_enable_walkme",
>     required=False,
>     help="Enable or disable guided tour",
>     action="store_const",
>     const="false"
> )
> ```
> 
> Much in the same way that `--disable-ca-trust` param works. Other than that tweak to the command flags I think the rest is good 👍

Incorporated the changes suggested and tested both scenarios in non-interactive mode again

durera (approved):

PR #1176, opened 2024-08-13
[patch] Add support for enabling special characters in ID/username field

RexMCM:
## Description
Adding support for the new special_character parameter

## Related Pull Requests
https://github.com/ibm-mas/python-devops/pull/20
https://github.com/ibm-mas/ansible-devops/pull/1412

## Related Issues

- [MASCORE-3417](https://jsw.ibm.com/browse/MASCORE-3417)

## How Has This Been Tested?
I conducted a full-person functional verification test (FVT) on the entire development branch, including the CLI.
image
 
Also, used CLI to install a new instance using these development branches to make sure the manual process will work correctly:
image
image


python/src/mas/cli/install/app.py
```diff
@@ -184,6 +184,10 @@ def configMongoDb(self) -> None:
         self.printH1("Configure MongoDb")
         self.promptForString("Install namespace", "mongodb_namespace", default="mongoce")
 
+    def configSpecialCharacters(self):
```
durera, line 188: What versions of MAS support this ... we should only show this prompt if the MAS version supports the option

python/src/mas/cli/install/app.py
```diff
@@ -856,7 +861,7 @@ def nonInteractiveMode(self) -> None:
                             self.fatalError(f"Unsupported format for {key} ({value}).  Expected string:int:int:boolean")
 
             # Arguments that we don't need to do anything with
-            elif key in ["accept_license", "dev_mode", "skip_pre_check", "skip_grafana_install", "no_confirm", "no_wait_for_pvc", "help"]:
+            elif key in ["accept_license", "dev_mode", "skip_pre_check", "skip_grafana_install" ,"no_confirm", "no_wait_for_pvc", "help"]:
```
durera: Typo?

durera (changes requested):
Additional changes are needed to ensure we have support for non-interactive install as well:

- the install command doc needs to be updated to reference `--allow-special-chars`
- argparser.py needs to be updated to support `--allow-special-chars`
- argbuilder.py needs to be updated to support `--allow-special-chars`
- app.py needs to be updated to support `--allow-special-chars`

python/src/mas/cli/install/app.py
RexMCM: This change was back-ported to all mas version

python/src/mas/cli/install/argBuilder.py
```diff
@@ -59,6 +59,7 @@ def buildCommand(self) -> str:
         command += f" --mas-instance-id {self.getParam('mas_instance_id')}"
         command += f" --mas-workspace-id {self.getParam('mas_workspace_id')}"
         command += f" --mas-workspace-name \"{self.getParam('mas_workspace_name')}\"{newline}"
+        command += f" --allow-special-chars \"{self.getParam('mas_special_characters')}\"{newline}"
```
durera: This needs to be in a conditional, otherwise it will be added even when the user didn't set it ... it also doesn't have a vaue ... e.g.  this is what you are adding atm:

`--allow-special-chars true` or `--allow-special-chars true`

But we need logic that says add ``--allow-special-chars` only if `self.getParam('mas_special_characters') == true` ... you'll find many examples of this kind of logic for flags that don't have a value elsewhere in the script.

python/src/mas/cli/install/argParser.py
```diff
@@ -75,7 +75,15 @@ def isValidFile(parser, arg) -> str:
...
-
+# MAS Special characters 
+# -----------------------------------------------------------------------------
+masSpecialCharacters = installArgParser.add_argument_group("Mas Special Characters")
+masSpecialCharacters.add_argument(
+    "--allow-special-chars",
+    dest="mas_special_characters",
+    required=False,
+    help="Allow special chars for users username/ID",
+)
```
durera, line 87: This should be a boolean ... e.g. using `action=store_true`, see `--eck-enable-logstash` for example.

durera (changes requested):

durera (approved):

PR #1215, opened 2024-08-28
[minor] Added HMAC to COS role

Vishal-patel-7:
As a part of https://jsw.ibm.com/browse/MAXMF-455 we added new environment variable USE_HMAC to create COS with HMAC support. There is PR https://github.com/ibm-mas/ansible-devops/pull/1407 in ansible-devops regarding this.


whitfiea (changes requested):
The ansible-devops role now allows the config scope to be set and the use_hmac to be set, which is all fine. But that means that for an env that has assist, mvi and manage installed we need the following set:
- objectstorage-cfg at system scope with use_hmac as false
- objectstorage-cfg at app/workspace(for manage) scope with use_hmac as true

I don't think the changes here allow the scope to be set or for more than one config/dependency to be created

Vishal-patel-7:
> The ansible-devops role now allows the config scope to be set and the use_hmac to be set, which is all fine. But that means that for an env that has assist, mvi and manage installed we need the following set:
> 
> * objectstorage-cfg at system scope with use_hmac as false
> * objectstorage-cfg at app/workspace(for manage) scope with use_hmac as true
> 
> I don't think the changes here allow the scope to be set or for more than one config/dependency to be created

I don't know how to handle this but the aim for the changes are to give new environment variable support for ansible-devops role. @chiragcsatani can you confirm what we need to do ? This is to fix fvt issue for manage but for assist & mvi we may have to revisit this code and ansible-devops role also.

tekton/src/pipelines/gitops/gitops-mas-deps.yml.j2
```diff
@@ -214,6 +218,8 @@ spec:
           value: $(params.ibmcloud_resourcegroup)
         - name: ibmcloud_apikey
           value: $(params.ibmcloud_apikey)
+        - name: use_hmac
+          value: $(params.use_hmac
```
whitfiea: missing bracket

whitfiea (changes requested):

tekton/src/pipelines/gitops/gitops-mas-deps.yml.j2
Vishal-patel-7: Corrected..

whitfiea (approved):

PR #1226, opened 2024-09-02
[minor] Support NFS storage on Fyre clusters

jcmaxisd:
This PR adds the following:

- Prompt/option to additional disks when provisioning Fyre cluster

- Prompt/option to set size of Fyre Quick Burn cluster as medium or large

- Prompt/option to set storage provisioner for Fyre (ocs or NFS)

- Additional options for ocs (setup additional disks in worker nodes) and NFS (setup image registry size)

Tests:

Various NFS instalations performed using Fyre Quick Burn provisioning, and also ODF installations using the same cluster type, both in interactive and non-interactive mode.


andrercm (dismissed):
LGTM!

durera (approved):

PR #1235, opened 2024-09-10
[patch] Add support for mirroring odf dependcies that are required for 4.15

IanBoden:
Just adds the mirror_odf option that was added into ansible-devops

Tested using branch of fvt-airgap based on mirror-images branch with extra odf option passed in called mirror-images-ib


whitfiea (commented):
@IanBoden can you add details of what testing has been done?

whitfiea (approved):

PR #1254, opened 2024-09-18
[patch] mas-devops 1.6.0

tomklapiscak:
This PR locks the version of mas-devops bundled into the CLI image to [1.6.0](https://github.com/ibm-mas/python-devops/releases/tag/1.6.0). Locking the mas-devops version feels like the right thing to do, otherwise updates to mas-devops will just happen incidentally in some other potentially entirely unrelated commit.


tomklapiscak:
Advised that we want to keep versions unlocked:

> ... otherwise it’s more complicated for working across the repos … eg we don’t lock version of ansible-devops either … the cli is always “use latest of our stuff”, we already control what goes into ansible-devops and coordinate with updates across the three repos … we want to avoid a scenario where a change in python-devops needs three prs everytime basically.

PR #1255, opened 2024-09-18
[minor] Add sanity tests for gitops

whitfiea:
Adds a sanity test framework for gitops that allows a set of tests to run in gitops when a flag ist set to run these tests. This allows the tests to be run on environments that will allow these type of tests (e.g. staging env).

The flag to use is `run_sanity_test` and has been tested in fvtsaas

Linked with gitops change https://github.com/ibm-mas/gitops/pull/195 

image- 2024-09-18 at 21 20 45


tomklapiscak (approved):
LGTM

PR #1285, opened 2024-10-04
[patch] Remove installplan from non-gitops pipeline

whitfiea:
The installplan vars were incorrectly added to the non-gitops pipelines, which was breaking the tekton install. Removing these and verified it can be installed.


sanjayprab (approved):

sanjayprab:
thanks Andrew for spotting this, missed to remove it

PR #1322, opened 2024-10-28
[patch] Update FVT Pipeline to run monitor test cases after manage DD

kisna13:
This PR is to execute the monitor test cases after the execution for core-mif-dd test case in manage. As some of the test cases in monitor when run with manage require data from manage.


durera (commented):
Has this change been tested somewhere already?

durera (approved):

PR #1333, opened 2024-11-06
[patch] Fix dro interactive mode

HariPalleti:
## Description:
DRO update is not doing anything in interactive mode

## Issue:
https://jsw.ibm.com/browse/MASCORE-4258

## Test:
Tested on the cluster with UDS to migrate to DRO in interactive mode


lokesh-sreedhara (dismissed):

python/src/mas/cli/update/app.py
```diff
@@ -437,21 +446,16 @@ def detectUDS(self) -> None:
...
-                            ])
-                            for storageClass in getStorageClasses(self.dynamicClient):
-                                print_formatted_text(HTML(f"<LightSlateGrey>  - {storageClass.metadata.name}</LightSlateGrey>"))
-                            self.promptForString("DRO storage class", "dro_storage_class", validator=StorageClassValidator())
+
+                        if self.getParam("dro_storage_class") is None or self.getParam("dro_storage_class") == "":
+                            self.selectDROStorageclass()
 
                 if self.getParam("dro_migration") == "true":
+                    self.showUDSUpdateNotice()
```
whitfiea: It seems that this notice will display even if UDS is not installed? That would be quite confussing

python/src/mas/cli/update/app.py
```diff
@@ -437,21 +446,16 @@ def detectUDS(self) -> None:
                         self.showUDSUpdateNotice()
                         self.fatalError(f"By choosing {self.getParam('mas_catalog_version')} you must provide the storage class to use for the migration to DRO using '--dro-storage-class' when using '--no-confirm'")
                     else:
-                        h.stop_and_persist(symbol=self.successIcon, text="IBM User Data Services needs to be migrated to IBM Data Reporter Operator")
-                        self.showUDSUpdateNotice()
```
whitfiea, line 441: Removing the notice here means we could just exit without much feedback to the user as to why

whitfiea (changes requested):

python/src/mas/cli/update/app.py
HariPalleti: Moved the notice to the else condition when UDS installed

python/src/mas/cli/update/app.py
HariPalleti: Added the notice here to show in the case of UDS installed

whitfiea (approved):

PR #1347, opened 2024-11-13
[patch] Fixes and updates for IBMCloud COS support

twhart:
Setup COS so it doesn't rely on the ibmcloud_apikey if the cos_apikey has been set. This allows for a less privileged apikey to be used. Renamed some variables to align more closely to how the rest of COS is used

Tested here:
https://dashboard.masdev.wiotp.sl.hursley.ibm.com/tests/pfvtcos

See also: MASCORE-4345


durera (changes requested):
the tekton stuff should continue to support both IBMCLOUD_XXX, with the COS_XXX env vars as an additional supported set (but in the CLI will will only use the new COS_XXX VERSION)

durera (approved):

PR #1352, opened 2024-11-15
[patch] MAXMOA-4297 - Mobile pipelines unique setups update

leo-miran:
Introducing new and unique setup for each of the mobile pipelines: fvt-mobile-pytest and fvt-mobile-testng.
Also enabling the response time param for all mobile tasks so that response time summary file is included in the test logs.

Successful test in fyre-evns screenshots:
- Pytest pipeline
image

- Testng pipeline
image


JonahLuckett (approved):
lgtm and tested in fyre-evns

PR #1357, opened 2024-11-18
[patch] fix civil setup dependency and mobile tasks reorder

leo-miran:
Fix missing setup task dependency for civil-bvt task and slight reorder of mobile tasks in order to increase execution efficiency.

image


JonahLuckett (approved):
Approved as a reorder of https://github.com/ibm-mas/cli/pull/1352 and deployed in fyre.

PR #1368, opened 2024-11-24
[patch] Clean up redundant files

durera:
- Various now-unused files were left behind after the conversion to the python-based install, update, and upgrade functions.
- Introduces concurrency controls into the CLI build (only one concurrent build per-branch)


whitfiea (changes requested):
the ibm-mas_devops.tar.gz needs to be removed

whitfiea (approved):

PR #1396, opened 2024-12-08
[patch] MAXMF-501 Export CommitId-Finaliser

chiragcsatani:
Export CommitId-Finaliser


image/cli/app-root/src/finalizer.py
```diff
@@ -350,9 +350,11 @@ def getcp4dCompsVersions():
                         if "mas.ibm.com/buildId" in deploymentDict["metadata"]["labels"]:
                             productBuildId = deploymentDict["metadata"]["labels"]["mas.ibm.com/buildId"]
                             productBuildNumber = deploymentDict["metadata"]["labels"]["mas.ibm.com/buildNumber"]
+                            productCommitId = deploymentDict["metadata"]["labels"]["mas.ibm.com/commitId"]
```
whitfiea: should we check if commitId is in the labels?

whitfiea (changes requested):

image/cli/app-root/src/finalizer.py
chiragcsatani: I have updated.

whitfiea (approved):

PR #1400, opened 2024-12-11
[minor] Add arm64 support

whitfiea:
Adds arm64 support for the container image so that the cli can be run on arm64 architecture. The build needs to be split as it was running out of space when attempting to build 3 container images in the same job.

Output of pulling this image on arm64 (note that the branch is called ajw-arm64 so the example below is pulling the image manifest and determining to pick arm64, rather than pulling ajw-arm64-arm64)

image- 2024-12-11 at 11 32 44@2x


durera (approved):
LGTM

PR #1456, opened 2025-02-05
[patch] sls_mongodb_cfg_file for custom mongodb namespace

sarthakSharma5:
## Fix for sls_mongodb_cfg_file path
Fixes a bug which causes install failure in SLS stage when using non-interactive method and a custom MongoDB namespace is provided in parameters.

Sets the parameter `sls_mongodb_cfg_file` to `/workspace/configs/mongo-{mongodb_namespace}.yml`

## Issue
By default, MAS installation fails when using the non-interactive method and opting for a custom MongoDB namespace.
This happens because the value for the parameter `sls_mongodb_cfg_file` remains at default, which is then used by the [mas_devops.sls](https://github.com/ibm-mas/ansible-devops/tree/master/ibm/mas_devops/roles/sls#sls_mongodb_cfg_file) role and causes a failure since the config file is not present at the path provided (which defaults to `/workspace/configs/mongo-mongoce.yml`).

To Replicate use `--mongodb-namespace "mongoce-$MAS_INSTANCE_ID"` with the non-interactive install command.

### Before Fix (failing SLS task during install)
image

##### - SLS task run
image

##### - FVT Runs for a MAS Installation (with IoT)
image


rawa-resul:
Could you add your condition after line 963 so that it is under the `#SLS` portion:

```bash
            # SLS
            elif key == "license_file":
                if value is not None and value != "":
                    self.slsLicenseFileLocal = value
                    self.setParam("sls_action", "install")
            elif key == "dedicated_sls":
                if value:
                    self.setParam("sls_namespace", f"mas-{self.args.mas_instance_id}-sls")
            # if mongodb_namespace is provided, set SLS MongoDB Configuration File path
            elif key == "mongodb_namespace":
                self.setParam("sls_mongodb_cfg_file", f"/workspace/configs/mongo-{value}.yml")
```

python/src/mas/cli/install/app.py
```diff
@@ -866,6 +866,10 @@ def nonInteractiveMode(self) -> None:
                 if value is not None:
                     self.setParam(key, value)
 
+                    # if mongodb_namespace is provided, set SLS MongoDB Configuration File path
```
rawa-resul: Could you add your condition after line 963 so that it is under the `#SLS` portion:

```bash
            # SLS
            elif key == "license_file":
                if value is not None and value != "":
                    self.slsLicenseFileLocal = value
                    self.setParam("sls_action", "install")
            elif key == "dedicated_sls":
                if value:
                    self.setParam("sls_namespace", f"mas-{self.args.mas_instance_id}-sls")
            elif key == "mongodb_namespace":
                self.setParam("sls_mongodb_cfg_file", f"/workspace/configs/mongo-{value}.yml")
```

rawa-resul (changes requested):

python/src/mas/cli/install/app.py
sarthakSharma5: For that to work, we would need to remove `mongodb_namespace` from [params.py#L74-L75](https://github.com/ibm-mas/cli/blob/slsmongocfg.ns/python/src/mas/cli/install/params.py#L74-L75) else it gets matched with the preceding condition (elif-block for `optionalParams`) within the `nonInteractiveMode` method at [app.py#L865](https://github.com/ibm-mas/cli/blob/slsmongocfg.ns/python/src/mas/cli/install/app.py#L865). Removing it from there might break other functions/methods.

sarthakSharma5: else we can have the elif-block for `optionalParams` shifted towards the very end, something like [this](https://github.com/ibm-mas/cli/pull/1456/commits/baf2e57f557c5389eff42502df654a41d9363a4c).
image

rawa-resul: I feel like since we are intercepting this argument it should not go straight to optionalParams anymore. We could remove 
```bash
    # MongoDb
    "mongodb_namespace",
```
in params.py and then in app.py just do:

```bash
            elif key == "mongodb_namespace":
                if value is not None and value != "":
                    self.setParam(key, value)
                    self.setParam("sls_mongodb_cfg_file", f"/workspace/configs/mongo-{value}.yml")
```

sarthakSharma5: Yaa, that works better. Let me do the change right away. Thanks for the tip!

rawa-resul: Thanks Sarthak

rawa-resul (approved):

PR #1467, opened 2025-02-13
[patch] Run watsonstudio cfg even if cp4d not installed locally for gitops

whitfiea:
Removes the requirement in the gitops pipeline that the cp4d wsl service is installed on the same cluster. This will allow the watsonstudio config to be set for cp4d that is either oncluster or offcluster.

when offcluster then the following secrets should be set in SecretsManager:

```
WATSON_STUDIO_SECRET=${ACCOUNT_ID}${SECRETS_KEY_SEPERATOR}${CLUSTER_ID}${SECRETS_KEY_SEPERATOR}${MAS_INSTANCE_ID}${SECRETS_KEY_SEPERATOR}wsl-cp4d

SECRET_KEY_WATSON_STUDIO_USERNAME=${WATSON_STUDIO_SECRET}#username
SECRET_KEY_WATSON_STUDIO_PASSWORD=${WATSON_STUDIO_SECRET}#password
SECRET_KEY_WATSON_STUDIO_URL=${WATSON_STUDIO_SECRET}#url
```

Also removes the mas_config_scope as this is actually being set as system within the task, and predict is deployed to use system scope

Adds in gitops-delete-jdbc-config-iot that was missing when IoT is removed.


maulik-modi22:
Can you check if it supports CA cert to have an additional layer of security?

whitfiea:
@maulik-modi22 there is no validation for Watson studio cfg in MAS Core. For Predict I don't know if server certificates are verified or not, I don't see an option to set it.

PR #1508, opened 2025-03-19
[minor] Change the order of Manage installation for authservice

joaopauloksn:
Given the code code for authservice (Manage foundation) work, we should now install Manage before installing any other app.

The personal fvt wth the changes are here: https://dashboard.masdev.wiotp.sl.hursley.ibm.com/tests/pfvtmfirst

To avoid compatibility issues with previous version we always have more than one runAfter for the applications that did not require Manage in the past. It will ensure that when manage is not installed (skipped) the application still proceeds with the installation.


whitfiea (changes requested):
Do you have the timings of how long it would take to install Manage Foundation. This change would add time to the deployment as previously, IoT, Monitor, Assist and MVI could be installed in parallel with Manage. With this change all apps now have to wait for Manage.

We will also need to ensure that https://ibm-mas.github.io/cli/reference/dependencies/ is updated with the new 9.1 topology at some point.

joaopauloksn:
> Do you have the timings of how long it would take to install Manage Foundation. This change would add time to the deployment as previously, IoT, Monitor, Assist and MVI could be installed in parallel with Manage. With this change all apps now have to wait for Manage.
> 
> We will also need to ensure that https://ibm-mas.github.io/cli/reference/dependencies/ is updated with the new 9.1 topology at some point.

@whitfiea , Installing Manage Foundation won’t be all that different from installing the full version. The only real difference is in the image download step. It might save about 10 minutes depending on the server network speed.

The whole idea behind Manage Foundation is to make it easier to upgrade to full Manage later on this is the reason why it is close to the full Manage version. 
Technically, the applications can still be installed in parallel, but we’re holding off because they won’t be functional without the authorization module anyway.

I'm running a new fvt core with Manage Foundation only and will post the results here by the end of the day.

joaopauloksn:
In the best case scenario, having Manage deployment holding other apps deployment will bring around 1 hour delay to the installation process:
image

In most cases, customer will need Manage in their environment anyway so this time will not really impact.

whitfiea (approved):

PR #1517, opened 2025-03-26
[patch] Add auth services into manage pipeline

RexMCM:
Adding new tests for authorization services in the FVT-Manage pipeline.

test result:
image


alequint:
@RexMCM I see pipeline is properly designed, but did not have a chance to run an fvt personal with that change in place? Do we have state of PipelineRun as a test evidence here?

alequint (changes requested):
PipelineRun test evidence

alequint (approved):
@RexMCM did run pipeline, test result was displayed and persisted in dashboard

PR #1556, opened 2025-04-23
[patch] Configure PVC for SNO mode

unnati-solanki-git:
https://jsw.ibm.com/browse/MASCORE-6326


unnati-solanki-git:
This is covered already

PR #1570, opened 2025-04-29
[minor] Support for Manage Foundation installation during an upgrade from MAS 9.0 without Manage to MAS 9.1

blechner:
Changes in this PR:
- In the scenario where MAS 9.0 without Manage is upgraded to MAS 9.1, the `mas upgrade` command will now install Manage Foundation. Added DB2 and Manage installation tekton tasks in `upgrade.yml.j2` for this specific scenario.
- The app-manage-upgrade task now runs before the other app upgrade tasks.
- When installing Manage Foundation, we need to set `mas_app_settings_aio_flag` param to false (both install and upgrade scenarios).
- Added an optional silentMode to configDb2, to make sure Db2 can be installed along with Manage Foundation in the upgrade scenario without showing any prompts.
- Added `--dev-mode` to `mas upgrade` so that the upgrade channel for Manage Foundation installation can be entered manually during tests. Artifactory username and token are also requested in dev mode.

Tested the following scenarios:
- Installation of MAS 9.0 without Manage and then upgrade to MAS 9.1 (Manage Foundation was installed)
- Installation of MAS 9.0 with Manage and then upgrade to MAS 9.1 (Manage 9.0 was upgraded to Full Manage 9.1)


alequint (dismissed):

alequint (approved):

alequint:
Looks good to me!

alequint:
Just changed PR title to include `minor` prefix

PR #1593, opened 2025-05-12
[patch] Updating CLI docs as rebranding Assist to Collaborate

mahima-shriwas-111:
As we are rebranding Assist to Collaborate, updating the document in application install section so that external will understand Assist is now labeled as Collaborate after logging in application.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": "^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2025-05-05T09:43:59Z",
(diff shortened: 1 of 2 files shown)


whitfiea (changes requested):
These are external docs so it needs to be written for an external point of view. We can't say "The plan is" or talk about "Label change"

mahima-shriwas-111:
> These are external docs so it needs to be written for an external point of view. We can't say "The plan is" or talk about "Label change"

Right, I have improved accordingly.

PR #1594, opened 2025-05-12
[doc] Updating CLI docs as rebranding Assist to Collaborate

mahima-shriwas-111:
As we are rebranding Assist to Collaborate, updating the document in application install section so that external will understand Assist is now labeled as Collaborate after logging in application.


docs/guides/install.md
```diff
@@ -154,6 +154,7 @@ The interactive install will guide you through a series of questioned designed t
       <ul>
         <li>Monitor is only available for install if IoT is selected</li>
         <li>Assist and Predict are only available for install if Monitor is selected</li>
+        <li>From now on Assist will be labeled as Collaborate. Although while installing the Assist you will see Assist right now but after logging in you will find Collaborate in place of Assist. Soon this is also gonna be improved and while installing also you will be able to see Collaborate.</li>
```
whitfiea: I am not sure this still reads right. I think it should be more like:
"From MAS 9.1 onwards, Assist will be rebranded as Collaborate in the MAS UI. It will still appear as Assist in the MAS CLI and within the OpenShift Cluster, but from the MAS UI it will appear as Collaborate."

whitfiea (changes requested):

docs/guides/install.md
mahima-shriwas-111: done

whitfiea (approved):

PR #1610, opened 2025-05-23
[minor] mas rollback

anilprajapatiibm:
Introducing mas rollback command


durera:
At this point the introduction rollback support is basically cancelled, so closing this PR

PR #1640, opened 2025-06-17
[patch] add db-scripts for facilities

terc1997:
**Description**
Maximo Real Estate and Facilities, previously TRIRIGA Application Suite, have shell scripts to configure external databases. These were previously available in https://github.com/IBM/tas-db-prep. Now, these scripts have been added to CLI image.


docs/guides/facilities-external-db.md
```diff
@@ -0,0 +1,16 @@
+Facilities External DB
```
sanjayprab, line 1: Needs to be referred as Maximo Real Estate and Facilities

sanjayprab (approved):

PR #1673, opened 2025-07-14
[major] Avoid usage of an additional variable to install Manage foundation

joaopauloksn:
## Issue
MASCORE-7612

## Description
We are removing the need of setting an specific IS_FULL_MANAGE variable when installing Manage Foundation.
This change will ensure that by default foundation is always installed if no components is set for Manage installation.

To keep compatibility with current behavior we are also change Manage oneclick playbook to always install Manage base e the env var is not set. The way to install foundation explicitly using the playbook is by setting the attribute as:
  --manage-components ""

In case someone is trying to install MAS 9.0.x with   --manage-components "", we flag an error as it is not allowed anymore.

## Test Results
I ran all sort of possible tests for this playbook including:
* Installed Foundation using both iteractive and non-iteractive mode;
* Instaled full manage using both iteractive and non iteractive mode;
* Upgraded 9.0 to 9.1 using MVI, installing foundation automatically

<hr/>


python/src/mas/cli/install/app.py
```diff
@@ -1143,6 +1137,10 @@ def nonInteractiveMode(self) -> None:
             self.validateCatalogSource()
             self.licensePrompt()
 
+        # Version before 9.1 cannot have empty components
```
durera, line 1140: as part of changing the default to remove health from Manage ...  (the reason we defaulted to health being installed was to make it easy for the predict usecase mainly), now the default is "", I feel like we need a variation of code like we have here that is saying "if predict is also being installed then make sure `"health" in self.getParam("mas_appws_components")` ... this seems like the right point to have that check too?  It's happening after all the apps have been configured, so maybe it's as simple as adding another `if` after this one?

joaopauloksn: ok updated:
image

durera (approved):

PR #1676, opened 2025-07-17
[patch] Pass allow/deny lists of test to run to cypress tests

jonquark:
Cypress tests currently ignore lists of testsuites to run/not run as it is not passed to their container.
This change passes the variable though (and uses a cypress specific task analogous to the pytest one which set many fewer variables etc).

It has been tested in fvt-personal:
https://dashboard.masdev.wiotp.sl.hursley.ibm.com/tests/pfvtlevuprf
(recent runs there use this change) - some of the tests fail but that is not to do with allow/deny lists (i.e. unrelatd to this change)


MaeveOReilly (approved):

unnati-solanki-git:
@jonquark Just wanted to take your attention to logic that allow us to run the test based on installed component. I think if we add a when condition like [this](https://github.com/ibm-mas/cli/blob/master/tekton/src/pipelines/taskdefs/fvt-manage/phase1.yml.j2#L44-L47) for [base-ui-userprofile](https://github.com/ibm-mas/cli/blob/master/tekton/src/pipelines/taskdefs/fvt-manage/phase5.yml.j2#L92) test then we do not need to maintain it through white list/black list.

jonquark:
@unnati-solanki-git that would stop this test running when we only have foundation - I'm arguing we need to have *more* tests including base-ui-userprofile running in foundation mode - so the change you suggest is the opposite to this one :)

jonquark:
(I've got a bunch of PRs in flight but the outcome should be that this is run in fvtcore where we only have foundation but you can control in which env(s) it runs (like you can for almost any other suite already with the allow deny lists))

sanjayprab (approved):

PR #1687, opened 2025-07-24
[patch] set to false aiservice var MAS_AIBROKER_SAAS

karol-czarnecki:
This PR is to set default value to false for MAS_AIBROKER_SAAS


durera:
If these changes are still needed please re-raise PR after resolving conflicts with the cleanup branch that has now been merged

PR #1693, opened 2025-07-29
[patch] Update FVT Pipeline to run monitor test cases in Tekton Monitor Pipeline

jvaneel-dwivedi:
The Monitor Logs were being dumped into ivt-manage pods instead of fvt-monitor.
Screenshot 2025-07-24 at 6 02 32 PM

We have fixed that by updating the values In https://github.com/ibm-mas/cli/blob/master/tekton/src/pipelines/fvt-monitor.yml.j2 

and removed the task from : https://github.com/ibm-mas/cli/blob/master/tekton/src/tasks/ivt/ivt-manage.yml.j2

We tested the same under the personal_fvt : https://dashboard.masdev.wiotp.sl.hursley.ibm.com/tests/pfvtjdibm

Currently we can see the logs in the desired suite(second Screenshot):
Screenshot 2025-07-29 at 6 31 50 PM


whitfiea (changes requested):
I thought the reason for having the monitor_fvt_with_manage run as ivt in manage was so that we make sure that the monitor fvt runs after manage has been installed, otherwise it runs too early. Is this not the case now?

jvaneel-dwivedi:
> I thought the reason for having the monitor_fvt_with_manage run as ivt in manage was so that we make sure that the monitor fvt runs after manage has been installed, otherwise it runs too early. Is this not the case now?

The Monitor test cases can be initiated as soon as Manage is installed; it is not necessary to wait for the completion of Manage IVT. Once Manage is up, we are good to proceed with running the Monitor tests.

unnati-solanki-git (approved):

whitfiea (approved):

PR #1699, opened 2025-08-01
[patch] Resolving Issue with Region Not Updating in ODH Files

Hardik-Prajapati-10:
This pull request addresses a bug where the region configuration was not being properly updated in the ODH files. The issue stemmed from the update logic not persisting the new region value as expected.

Changes include:
Corrected the logic responsible for updating the region value in ODH


.secrets.baseline
durera, line 1: Need to resolve the conflicts in this file before this PR can be merged

durera (changes requested):

.secrets.baseline
Hardik-Prajapati-10: thanks @durera , 
conflict resolved

image/cli/mascli/functions/gitops_odh
```diff
@@ -67,7 +67,7 @@ Storage:
       --storage-host
       --storage-port
       --storage-ssl
-      --storage-region
+      --mas-aibroker-storage-region
```
durera: This looks wrong based on all the other params not having the `--mas-aibroker-` prefix, can you double check this is intended?

image/cli/mascli/functions/gitops_odh
```diff
@@ -315,7 +315,7 @@ function gitops_odh_noninteractive() {
       --storage-ssl)
         export MAS_AIBROKER_STORAGE_SSL=$1 && shift
         ;;
-      --storage-region)
+      --mas-aibroker-storage-region)
```
durera: adding the --mas-airboker prefix seem unnecessary, we don't have it for any other the other similar named params?

durera (changes requested):

image/cli/mascli/functions/gitops_odh
Hardik-Prajapati-10: --mas-airboker prefix removed

image/cli/mascli/functions/gitops_odh
Hardik-Prajapati-10: --mas-airboker prefix removed

durera (approved):

PR #1706, opened 2025-08-04
[patch] add validator for uppercase language for manage installations

### LINKED ISSUE #1445: MAS CLI installation needs case validation for language selection
### Feature and motivation

During the MAS / Manage installation process using the CLI, one of the steps is to provide secondary / additional languages. If you provide an additional language in a lower case (for example, 'fr' instead of 'FR') it will continue without returning a warning/error message. However, the installation will still fail later due to the lowercase language code.

The suggestion is for the CLI to either immediately prompt an error if a lowercase language code is provided, or have the installer automatically convert the lowercase language code into its proper capitalized form.

### Usage example

This enhancement would avoid a lot of situations where time is lost due to failed installations because the error would be caught much earlier.

terc1997:
## Issue
#1445 
#1529 

## Description
Created a validator to inform the user when the language in Manage is inputed as lower case for interactive mode. For non-interactive mode, automatically update language value to upper case.

## Test Results
Steps:
### Interactive mode:
- All lower case
image

- First letter upper case
image

- Last letter upper case
image

Same approach was applied to secondary language. See screenshots below:
image

image

image

### Non Interactive mode:
image

<hr/>


python/src/mas/cli/install/settings/manageSettings.py
```diff
@@ -224,15 +225,15 @@ def manageSettingsLanguages(self) -> None:
...
+        self.promptForString("Base language", "mas_app_settings_base_lang", default="EN", validator=UpperCaseValidator())
 
         self.printDescription([
             f"Define the additional languages to be configured in Maximo {self.manageAppName}. provide a comma-separated list of supported languages codes, for example: 'JA,DE,AR'",
             "A complete list of available language codes is available online:",
             "    <Orange><u>https://www.ibm.com/docs/en/mas-cd/mhmpmh-and-p-u/continuous-delivery?topic=deploy-language-support</u></Orange>"
         ])
 
-        self.promptForString("Secondary languages", "mas_app_settings_secondary_langs")
+        self.promptForString("Secondary languages", "mas_app_settings_secondary_langs", validator=UpperCaseValidator())
```
durera: Rather than validate the case on entry (forcing a user to delete/correct), can we apply uppercase automatically? ... this way the inputs can be any case and we can `.upper()` it after they've entered the string.

```
self.promptForString("Secondary languages", "mas_app_settings_secondary_langs")
# Manage only understands UPPERCASE language strings, so convert the input to uppercase
self.params("mas_app_settings_secondary_langs") = self.params("mas_app_settings_secondary_langs").upper()
```

This would be consistent with how you handled this for the non-interactive install .. we don't require the user to  provide the values in the correct case, we just silently fix it for them.

terc1997: Sure. I'll change accordingly.

durera:
Moved this into draft state pending the changes that were discussed in Slack around moveing from validator to selection list .. please move out of draft when ready for re-review

terc1997:
> Moved this into draft state pending the changes that were discussed in Slack around moveing from validator to selection list .. please move out of draft when ready for re-review

Couple changes from our last conversation. 
- Created `promptForListMultiSelect` since the secondary language can be an array
- Added validator for `promptForListSelect`, `promptForInt` (as dependency of `promptForListSelect`) and `promptForListMultiSelect`, because when you pass an out of index value it throws an error that makes you add all arguments all over again (which happened accidentally with me a couple times). Now, when you type an option of out index, it asks you to put it again.

### Tests
#### promptForListSelect()
- Out of index
image

- Acceptable value
image

#### promptForListMultiSelect()
- Out of index
image

- Acceptable value
image

### Summary
image

Let me know if I didn't overcomplicate the code. And if you agree with the stateful validator, we can add to other lists that we have in the code as a guardrail.

python/src/mas/cli/install/settings/manageSettings.py
```diff
@@ -224,15 +226,16 @@ def manageSettingsLanguages(self) -> None:
         self.printDescription([
             f"Define the base language for Maximo {self.manageAppName}"
         ])
-        self.promptForString("Base language", "mas_app_settings_base_lang", default="EN")
+        self.manageLanguageDisplay()
+        self.promptForListSelect("Base language", self.supportedLanguages, "mas_app_settings_base_lang", validator=OutOfIndex(self.supportedLanguages))
 
         self.printDescription([
-            f"Define the additional languages to be configured in Maximo {self.manageAppName}. provide a comma-separated list of supported languages codes, for example: 'JA,DE,AR'",
+            f"Define the additional languages to be configured in Maximo {self.manageAppName}. Provide a comma-separated list of the supported languages indexes, for example: '2,10,23'",
```
durera: Is there no way to have the user just choose the list of languages without having to translate that list into numbers, it feels a bit clunky having to enter numbers rather than the codes directly .... 

e.g.

```
validlangs = ["EN", "DE", "FR"]
input = "EN,FR"

## validate logic
multiEnumValidator(inputString: str, validChoicesList: list):
  inputAsList = inputstring.replace(" ", "").split(",")
  valid = all(item in validChoicesList for item in inputAsList)  # True if all items in inputAsList are contained in validChoicesList

```

durera (approved):

PR #1715, opened 2025-08-07
[patch] add ocp_ingress to overwrite ingress domain

### LINKED ISSUE #1734: Generated non-interactive command not including the --domain flag
### MAS CLI version

15.1.7

### CLI function used

install

### What happened?

If you set the custom domain during an interactive install, the non-interactive command that is generated later does not include the --domain flag.

### Relevant log output

```shell
mas_domain is being set to a string:
http://github.com/ibm-mas/cli/blob/aac19aab846139985b5a251c326b9bb1b69425b8/python/src/mas/cli/install/app.py#L507

The code should add the --domain flag if the configured custom domain (mas_domain) is set to true:
https://github.com/ibm-mas/cli/blob/master/python/src/mas/cli/install/argBuilder.py#L101

Other flags that use strings are checking if the string is empty, not if it is true/false. Not sure if this needs to be the case here or not.
```

terc1997:
## Issue
https://github.com/ibm-mas/ansible-devops/issues/1231
#1734 

## Description
Add an overwrite variable to OCP ingress domain.

Add feats:
- Fix `mas_domain` and `dns_provider` bug in argument builder.

## Test

This test verifies how CLI loads the parameter `ocp_ingress` and populates in the Tekton tasks.

#### Interactive mode
- Overwrite `ocp_ingress`
image
image

- Don't overwrite `ocp_ingress`
image
image
image

#### Non interactive mode
- Overwrite `ocp_ingress`
image
image
image

- Don't overwrite `ocp_ingress`
image
image
image

### Tekton
- `mas-devops-suite-dns` Task
image

- `install` Pipeline 
image

<hr/>


python/src/mas/cli/install/app.py
```diff
@@ -505,6 +505,8 @@ def configDNSAndCerts(self):
                 configureDomain = self.yesOrNo('Configure custom domain')
```
durera: Can we add a `printDescription` before this `Configure custom domain` prompt to explain what the mas top-level domain and the ingress domain overwrite do .. the former is pretty self-explanatory and shouldn't need much explanation, but I think for many it won't be immediately obvious what `Overwrite Ingress Domain?` means ... I'd have to look it up to be honest, even though I vaguely remember reviewing an ansible-devops PR about it not that long ago, so I think that's a sign we need to offer up some context in the interactive prompts.

durera (changes requested):

python/src/mas/cli/install/app.py
terc1997: ACK

durera (approved):

PR #1718, opened 2025-08-11
[patch] Fix aibroker/aiservice related issue

Bhautik-Vala:
this changes is for solving some issue in interactive mode of aiservice installation


tekton/src/pipelines/taskdefs/dependencies/db2.yml.j2
```diff
@@ -37,7 +37,7 @@
       value: $(params.db2_timezone)
 {% endif %}
 {% if suffix == "aiservice" %}
-      value: aiservice-$(params.aiservice_instance_id)
```
durera: I think the change needs to be the other way around .. we need the aiservice db2 instance to carry the aiservice instance id on it, otherwise if we have two aiservice instances the db instances will clash.  What things are expecting the db2 instance id to just be `aiservice` instead of `aiservice-{aiserviceinstanceid}`

durera (changes requested):

durera (approved):

PR #1720, opened 2025-08-11
[DOC Update] Update v9-250731-amd64.md

vsolasa:
Its a known issue and if we run mas update without the option --mongodb-v7-upgrade, it will fail. Hence documented it on release known issues.


durera:
This doesn't seem right to me, following up in Slack ... that flag should only be necessary if the update will trigger an upgrade from mongo v6 to v7, I would have thought our testing would catch an issue if not setting that flag prevented update otherwise, and in the case where update will trigger a mongo upgrade this is simply working as designed, and doesn't need a known issue.

durera:
As previously discussed, this was a bug in the implementation that has now been fixed.  The upgrade flags are only required for major version changes in MongoDb

Issue #1542, opened 2025-04-09
Provide a warning if MAS CLI must-gather is running without being logged into a cluster


BrendanFishback:
### Feature and motivation

Currently, when a MAS CLI must-gather is run without logging into a OpenShift cluster, the must-gather will still execute and attempt to collect data. The collection logs will be full of messages like this:
```
error: Missing or incomplete configuration info.  Please point to an existing, complete config file:

  1. Via the command-line flag --kubeconfig
  2. Via the KUBECONFIG environment variable
  3. In your home directory as ~/.kube/config
```

While to someone experienced with OpenShift this may be an obvious sign the must-gather did not collect the data properly, it is not immediately apparent to those with less exposure to OpenShift. It would be beneficial if the must-gather script itself would provide a warning message to the end user that the must-gather did not execute properly due to not being logged in to a cluster instead.

### Usage example

This would save time from having to request data multiple times from clients with less experience with OpenShift.

PR #1724, opened 2025-08-14
[patch] add cluster verification in must-gather

nehasangwai-ibm:
[Provide a warning if MAS CLI must-gather is running without being logged into a cluster](https://github.com/ibm-mas/cli/issues/1542)

Modified the script as per the conditions. Currently, when a MAS CLI must-gather is run without logging into a OpenShift cluster, the must-gather will still execute and attempt to collect data. The collection logs will be full of messages like this:

error: Missing or incomplete configuration info.  Please point to an existing, complete config file:

  1. Via the command-line flag --kubeconfig
  2. Via the KUBECONFIG environment variable
  3. In your home directory as ~/.kube/config
While to someone experienced with OpenShift this may be an obvious sign the must-gather did not collect the data properly, it is not immediately apparent to those with less exposure to OpenShift. It would be beneficial if the must-gather script itself would provide a warning message to the end user that the must-gather did not execute properly due to not being logged in to a cluster instead.

Attaching snippets with & without the oc login

Screenshot 2025-08-13 at 10 22 13 PM
Screenshot 2025-08-13 at 10 26 19 PM

### DIFF
--- a/image/cli/usr/bin/gather
+++ b/image/cli/usr/bin/gather
@@ -2,6 +2,26 @@
 
 # https://github.com/openshift/enhancements/blob/89bbe226db20573f8f489eacb558a9f011072737/enhancements/oc/must-gather.md#must-gather-images
 
+# Check if the user is logged into an OpenShift cluster
+function check_cluster_connection() {
+  echo "Checking connection to OpenShift cluster..."
+  if ! oc whoami &>/dev/null; then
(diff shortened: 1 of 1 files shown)


terc1997 (commented):
LGTM. Your tests were done invoking `./script.sh`. Can you invoke `mas must-gather` and attach the screenshot as well ? Just to verify the workflow that the user follows.

durera (approved):

PR #1749, opened 2025-08-28
[patch] Add validation on bucket prefix

jasmin-30:
1. Add validation on bucket prefix 
2. Remove prompt for tenant entitlement type from ai service interactive install mode.

### DIFF
--- a/python/src/mas/cli/aiservice/install/app.py
+++ b/python/src/mas/cli/aiservice/install/app.py
@@ -42,7 +42,8 @@ from ...install.settings.additionalConfigs import AdditionalConfigsMixin
 
 from mas.cli.validators import (
     InstanceIDFormatValidator,
-    StorageClassValidator
+    StorageClassValidator,
+    BucketPrefixValidator
 )
 
 from mas.devops.ocp import createNamespace, getStorageClasses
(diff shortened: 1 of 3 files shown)


python/src/mas/cli/aiservice/install/app.py
```diff
@@ -549,7 +555,7 @@ def aiServiceTenantSettings(self) -> None:
 
         today = datetime.today()
         oneyear = datetime.today() + relativedelta(years=1)
-        self.promptForString("Tenant entitlement type", "tenant_entitlement_type", default="standard")
+        self.setParam("tenant_entitlement_type", "standard")
```
terc1997, line 558: What are the other types of `tenant_entitlement_type` ? In the ansible-devops, you can load the value from `AISERVICE_TENANT_ENTITLEMENT_TYPE` env var. If it's going to be always `standard`, it's better not set and let ansible-devops handle it and in the future just hard-code and remove the reference from `AISERVICE_TENANT_ENTITLEMENT_TYPE` also.

python/src/mas/cli/validators.py
```diff
@@ -149,3 +149,14 @@ def validate(self, document):
...
         except JSONDecodeError:
             raise (ValidationError(message='Enter a valid JSON', cursor_position=len(inputJson)))
+
+
+class BucketPrefixValidator(Validator):
+    def validate(self, document):
+        """
+        Validate Bucket prefix length
+        """
+        instanceId = document.text
```
terc1997: Follow clean code standard and name accordingly. For example, `bucketPrefix`.

python/src/mas/cli/validators.py
```diff
@@ -149,3 +149,14 @@ def validate(self, document):
...
+
+        if not match(r"^.{1,4}$", instanceId):
```
terc1997: Why not use simply `if len(instanceId) <= 4 and len(instanceId) > 0:` and rely on regex ?

terc1997 (changes requested):

python/src/mas/cli/aiservice/install/app.py
jasmin-30: We have multiple entitlement_types such as standard, essentials, premium. These are for SaaS deployments. We kept the option to opt for other entitlement type for dev environment when using cli non-interactive mode. but for production if customer install using cli. It will be set to standard by default.

python/src/mas/cli/validators.py
jasmin-30: I just followed WorkspaceNameFormatValidator which also checks for length and used the regex.

python/src/mas/cli/validators.py
jasmin-30: Updated

durera (approved):

durera (approved):

PR #1764, opened 2025-09-03
[patch] set mas_app_cfg_retries in cli

terc1997:
**Description**
The new installer for MAS in Techzone leverages the CLI and sometimes the default retries are not enough to complete the full deployment. To avoid that, I've added the parameter `mas_app_cfg_retries` into CLI that is already present in our Ansible collection to extend the retries.

**Ansible Reference**
https://github.com/ibm-mas/ansible-devops/blob/b3ff6534243ccaaa1387faea0c121ad49d699154/ibm/mas_devops/roles/suite_app_config/vars/manage.yml#L8


durera (commented):
I don't think we should directly expose this setting at all.  If the timeouts we have in the ansible roles per app are not long enough for low performance clusters then we should fix them.

This isn't a problem a customer should have to worry about at all, or be exposed to.

Which applications are timing out in techzone?  Let's fix the problem at the source ... in the ansible collection we should reconfigure that applications settings to allow longer.  

e.g if the problem is monitor then we need to adjust https://github.com/ibm-mas/ansible-devops/blob/master/ibm/mas_devops/roles/suite_app_config/vars/monitor.yml

```
mas_app_cfg_delay: "{{ lookup('env', 'MAS_APP_CFG_DELAY') | default(120, true)}}"
mas_app_cfg_retries: "{{ lookup('env', 'MAS_APP_CFG_RETRIES') | default(30, true)}}"
```

We should be treating mas_app_cfg_delay and mas_app_cfg_retries as an internal variable, that if a customer does need to reconfigure it we have gone wrong, our timeouts should be tailored for low performance clusters.  When we started using Fyre we had to increase all the timeouts because we were used to the higher performance of ROKS clusters, now we are seeing that techzone is even worse performance than Fyre we can use that the yardstick for what these timeouts should be set to.

PR #1771, opened 2025-09-09
[patch] added CLI commands for space_id deployment_id for watsonX

byalif:
MASAIB-1362


python/src/mas/cli/aiservice/install/app.py
```diff
@@ -178,6 +178,11 @@ def nonInteractiveMode(self) -> None:
 
         self.setDB2DefaultSettings()
 
+        if not hasattr(self.args, "aiservice_watsonxai_deployment_id") or self.args.aiservice_watsonxai_deployment_id is None:
+            self.setParam("aiservice_watsonxai_deployment_id", "default-deployment-id")
+        if not hasattr(self.args, "aiservice_watsonxai_space_id") or self.args.aiservice_watsonxai_space_id is None:
+            self.setParam("aiservice_watsonxai_space_id", "default-space-id")
```
durera: Do these values (default-deployment-id, and default-space-id) have some special meaning, that triggers some special code path in the ansible code?

byalif: actually we can remove all 4 of these lines. I added these initially because the parameters were undefined (I used a hyphen instead of underscore). But i fixed this in argParser.py and forgot to remove these lines.

durera (approved):

PR #1775, opened 2025-09-11
[patch] fix AI Service install with existing MongoDB

boonware:
## Issue
MASAIB-1380

## Description
* Fixes a bug in the interactive AI Service install path when the user selects to re-use an existing MongoDB installation. The method `selectLocalConfigDir` was missing from the AI Service installer class; it has now been moved to the base app class for all installer classes.
* Related to the above, the MongoDB code path expects the installer class to include the mixin `ConfigGeneratorMixin`.

## Test
* Two test scenarios have been implemented for the AI Service `install` method, one each for the interactive and non-interactive modes. These are long and ugly tests but are the best we can do right now (they are a starting point) since the code is not decoupled in a way that would allow easily unit testing the various components. The tests mock external calls and verify that the code path does not error out.
* It would be too cumbersome to write tests for every possible input parameter scenario with the code in its current form. Importantly, the interactive test case includes the code path that uncovered the bug above.
* Future work should look at generating all permutations of the input parameters and passing these to a set of parameterized tests to increase scenario coverage.
* Mocking the `prompt` function required that the message parameter was passed as a keyword parameter, so calls to this function have been updated in various places.


Bhautik-Vala (dismissed):

karol-czarnecki (dismissed):

python/report/coverage.xml
durera, line 1: Can we delete the report files and add `report/` to .gitignore so these are not checked in.

durera (changes requested):

python/report/coverage.xml
boonware: Done. My mistake, the report stuff wasn't meant to be in this branch.

durera (approved):

Issue #1636, opened 2025-06-13
mas cli (latest) mangles the custom database settings provided during install


sunilkumaraar:
### MAS CLI version

13.25.1

### CLI function used

install

### What happened?

mas cli (latest) mangles the custom database settings provided during install.
Although the cli accepts the schema, tablespace, indexspace values.
The actual considered values are schema, username, tablespace, indexspace.

Due to the above mismatch, the value passed for "schema" is taken by "username" and schema value considered is wrong.

Here are the sequence of events leading to the issue reported.

[mascli_13.25.1_custom_db_settings.txt](https://github.com/user-attachments/files/20721957/mascli_13.25.1_custom_db_settings.txt)

**** running mas cli latest (13.25.1)**

[root@host ~]# podman run --name masclin --rm -ti -v /images:/images:z -v ~:/mnt/home:z --pull always [quay.io/ibmmas/cli](https://quay.io/ibmmas/cli)
Trying to pull quay.io/ibmmas/cli:latest...
Getting image source signatures
...
Writing manifest to image destination
**IBM Maximo Application Suite CLI Container v13.25.1**

**** input provided**

17.3) Maximo Manage Settings - Database
Customise the schema, tablespace, indexspace, and encryption settings used by Manage
Customize database settings? [y/n] y
**Schema PRIMARY**
Tablespace PRIMARY
Indexspace dbo
Customize database encryption settings? [y/n] n

**** input considered for install are incorrect 

[mascli_13.25.1_custom_db_settings.txt](https://github.com/user-attachments/files/20721886/mascli_13.25.1_custom_db_settings.txt)

(confirmation)**

25.6) IBM Maximo Application Suite Applications
  IoT ..................................... Do Not Install
  Monitor ................................. Do Not Install
  Manage .................................. 9.0.x
...
  + Database Settings
    **+ Schema .............................. dbo
    + Username ............................ PRIMARY**
    + Tablespace .......................... PRIMARY
    + Indexspace .......................... dbo

However the re-usable command generated seems correct.
**** non-interactive command generated**

mas install --mas-catalog-version v9-250501-amd64 --ibm-entitlement-key $IBM_ENTITLEMENT_KEY \
  --mas-channel 9.0.x --mas-instance-id mas1 --mas-workspace-id mas1 --mas-workspace-name "mas1" \
...
  --manage-channel "9.0.x" \
  --is-full-manage "true" \
  --manage-jdbc "workspace-application" \
[...]

PR #1790, opened 2025-09-19
[patch] Fix db schema and username mismatch in summariser

kisna13:
There was a mismatch while displaying the summary before initiating the mas install
Fix for this issue -> https://github.com/ibm-mas/cli/issues/1636


terc1997 (approved):
LGTM

PR #1796, opened 2025-09-25
[patch] Remove check for slack-url and icd when creating aws secret for DB2 backups

sachbalag:
When creating the AWS secret containing the S3 bucket info for DB2U backup DB2_BACKUP_NOTIFY_SLACK_URL and ICD_AUTH_KEY were required. For FedRamp Slack and ICD services are not available inside the boundary hence we don't need that check.

Also removed the validation as well.


whitfiea (changes requested):
The code needs to continue to work in the commercial env where the icd and slack notificiations are required and do need to be set in SM. I believe the easiest way would be to introduce a new option that can be set on this function. It should default to the commerical (current) way of working so that we don't break anything, but if it is set then we bypass the setting and checking of these secrets. The `    sm_verify_secret_exists $SECRET_NAME_ICD_AUTH_KEY "icd_auth_key"` would also need to be bypassed.

The option could be called something like `BACKUP_NOTIFICATION` with a true/false value maybe?

sachbalag:
@whitfiea Here are the revised changes. Couple of questions ...

1) I'm assuming the new option needs to be documented? If so should it be in the gitops_db2u_database_help() function?
2) I wasn't sure where this function, gitops_db2u_database() is ultimately called from for FedRamp? I beleive we need to make sure SKIP_BACKUP_NOTIFICATION is set to true there?

image/cli/mascli/functions/gitops_db2u_database
```diff
@@ -513,6 +513,10 @@ DB2_WORKLOAD: '${DB2_WORKLOAD}'"
   if [[ -z $AUTO_BACKUP ]]; then
     export AUTO_BACKUP=false
   fi
+  # Option to track icd and slack notifications, and if they need to be set in secret manager. The default behavior for commercial SaaS is true.
```
whitfiea: > The default behavior for commercial SaaS is true
shouldn't that be false?

whitfiea (changes requested):
Yes the gitops_db2u_database_help needs to be updated, as well as the gitops_db2u_database_noninteractive to accept it as a parameter

image/cli/mascli/functions/gitops_db2u_database
sachbalag: Yes, that was a typo.

sachbalag: @whitfiea please review again.

whitfiea (approved):

PR #1811, opened 2025-10-02
[minor] Add Debug function supporting coredump collection

mattlrx:
java coredumps are created on the nodes and can be difficult to collect. this pull request adds a dedug command to generate and collect coredumps as well as a command to generate javacores at regular interval and collect them (helpful to troubleshoo performance issue)

to generate and collect the coredump, user will specify what liberty pod needs to be selected.
the function will exec to that pod(s) and generate the coredump, then a debug pod will be started on the node where the liberty pod was running and the coredump will be collected (using oc rsyc) .


terc1997 (dismissed):
Very cool feature. It will greatly add value to our CLI. However, since the CLI is a tool used across all apps under MAS, it needs to be application agnostic. Therefore, all Manage specific implementations need to be generalized since other applications use Liberty as well. So,
- Remove Manage specific implementations
- The debug namespace can be named just `mas-<instance id>-debug`
- Add a feature to clean up after debug is done.

Also, attach some evidences of testing.

mattlrx:
@terc1997 thank you for the comment, I have updated accordingly, also attaching test file with output from the commands tested.


[test.txt](https://github.com/user-attachments/files/22687182/test.txt)

durera (approved):

PR #1816, opened 2025-10-03
[patch] Updating 92 fvt-ibm-mas to run FVTs post manage installation

joaopauloksn:
With the centralization of user data in the Manage database, we need to refactor the Core test suite. Most of these tests rely on user creation, and with the removal of the User collection from MongoDB, only the superuser remains available by default.
To address this, we propose introducing a new environment variable: is_post_92. This variable will be calculated based on mas_channel information. If mas_channel information is not provided it will default to false.
This approach allows us to maintain compatibility with both pre- and post-9.2 versions. Once support for versions prior to 9.2 is deprecated, we can remove the flag and make the new behavior the default.
The changes have been tested with is_post_92 set to both false and true, and the results are consistent with expectations.

image

Results of my personal fvt cluster
image


durera (commented):
Rather than have a new param (and potential to set it wrong), can we use a generated result ... e.g.  a new stage in the pipeline named something like `config`

```
  results:
    - name: is-post-92
      description: Use post-9.2 FVT sequencing
  steps:
    - name: xxx
      script: |
        #!/bin/bash
        # logic to determine is params.mas_channel >= 9.2
        # resulting in IS_POST_92=true|false

        echo -n "$IS_POST_92" > $(results.is-post-92.path)
```

and then in the places we need this we can reference:

```
$(tasts.config.results.is-post-92)
```

@unnati-solanki-git already has something like this in place for determining the manage components that should be tested IIRC, we also have examples in the build system tekton files.

joaopauloksn:
> Rather than have a new param (and potential to set it wrong), can we use a generated result ... e.g. a new stage in the pipeline named something like `config`
> 
> ```
>   results:
>     - name: is-post-92
>       description: Use post-9.2 FVT sequencing
>   steps:
>     - name: xxx
>       script: |
>         #!/bin/bash
>         # logic to determine is params.mas_channel >= 9.2
>         # resulting in IS_POST_92=true|false
> 
>         echo -n "$IS_POST_92" > $(results.is-post-92.path)
> ```
> 
> and then in the places we need this we can reference:
> 
> ```
> $(tasts.config.results.is-post-92)
> ```
> 
> @unnati-solanki-git already has something like this in place for determining the manage components that should be tested IIRC, we also have examples in the build system tekton files.

@durera changes implemented and tested using fvt-stable template in a personal fvt. Can you please take a look?

blechner (approved):
LGTM

durera (approved):

PR #1836, opened 2025-10-14
[patch] oscap scanning and hardening images 

HariPalleti:
## Description:
To satisfy FedRAMP's requirements for container vulnerability MAS must leverage Hardened Images. 
OSCAP scan report will be generated and the report will be uploaded to artifactory which is used for CI evidence

## Issue:
- https://jsw.ibm.com/browse/MASCORE-9827

## Tests:

I have deployed MAS instanc in a development cluster after hardening the images


build/bin/.functions.sh
```diff
@@ -93,3 +93,9 @@ function artifactory_upload() {
   echo "Uploading $1 to $2"
   curl -H "Authorization:Bearer $ARTIFACTORY_TOKEN"  -H "X-Checksum-Md5: $md5Value" -H "X-Checksum-Sha1: $sha1Value" -T $1 $2 || exit 1
 }
+
+# install oscap tools
+function install_oscap() {
+  sudo apt-get update
```
terc1997: Since in https://github.com/ibm-mas/cli/blob/c442bda8c48e71187eae93581e5335d4860da4c8/build/bin/docker-build.sh#L62 you already have the update embedded, you don't need to perform again in here. I believe it would be better to remove the oscap tools installations from here and perform them directly in https://github.com/ibm-mas/cli/blob/508d227415a65a4a3ad558ddab59aadb56c68a9a/build/bin/docker-build.sh#L93 by calling `sudo apt-get install -y openscap-scanner libopenscap25`

build/bin/docker-build.sh
```diff
@@ -78,3 +82,32 @@ docker buildx build --progress plain \
...
+    else
+      sudo $DIR/oscap-docker.sh $REPOSITORY:$DOCKER_TAG-$TARGET_PLATFORM xccdf eval --report $OSCAP_DIR/$image_name-report.html --results $OSCAP_DIR/$image_name-results.xml --profile stig $CONFIG_DIR/oscap/${SCAP_DATA_STREAM}.xml
+    fi
+    sudo oscap xccdf generate fix --fix-type bash --output $OSCAP_DIR/$image_name-remediation.sh --result-id xccdf_org.open-scap_testresult_xccdf_org.ssgproject.content_profile_stig $OSCAP_DIR/$image_name-results.xml
+    
+    # Upload the results to Artifactory    
+    artifactory_upload $OSCAP_DIR/$image_name-report.html $ARTIFACTORY_GENERIC_RELEASE_URL/ibm-mas/$image_name/$DOCKER_TAG/$image_name-report.html
+    artifactory_upload $OSCAP_DIR/$image_name-results.xml $ARTIFACTORY_GENERIC_RELEASE_URL/ibm-mas/$image_name/$DOCKER_TAG/$image_name-results.xml
+    # To do: Getting permissions denied while uploading the $image_name-remediation.sh to artifactory since the file has readonly permissions only to root user
+    #artifactory_upload $OSCAP_DIR/$image_name-remediation.sh $ARTIFACTORY_GENERIC_RELEASE_URL/maximoappsuite/$image_name/$DOCKER_TAG/$image_name-remediation.sh
```
terc1997: Are you going to fix this PR ? If not, keep only the TODO and already create a Jira ticket to do it.

build/bin/oscap-docker.sh
```diff
@@ -0,0 +1,100 @@
...
+    echo
+    echo "Compliance scan of Docker container:"
+    echo "$ sudo oscap-docker [--oscap=<OSCAP_BINARY>] CONTAINER_NAME OSCAP_ARGUMENT [OSCAP_ARGUMENT...]"
+    echo
+    echo "See \`man oscap\` to learn more about semantics of OSCAP_ARGUMENT options."
+}
+
+OSCAP_BINARY=oscap
+
+if [ $# -lt 1 ]; then
```
terc1997, line 36: Instead of using `if`, change it to switch to be in the same format as other arguments in the CLI and keep the structure more clean. Example -> https://github.com/ibm-mas/cli/blob/508d227415a65a4a3ad558ddab59aadb56c68a9a/image/cli/mascli/mas#L106-L658

terc1997 (changes requested):

build/bin/docker-build.sh
HariPalleti: I have removed the comments from here. Since this file is not needed.

build/bin/oscap-docker.sh
HariPalleti: Switch statements might be more complicated for this scenario. Better to have them in if else ? Let me know your thoughts.
```
case $# in
    0)
        invalid "No arguments provided."
        ;;
    *)
        case "$1" in
            -h|--help)
                usage
                exit 0
                ;;
            --oscap=*)
                if [ $# -gt 2 ]; then
                    OSCAP_BINARY=${1#"--oscap="}
                    shift
                else
                    invalid "Invalid arguments provided."
                fi
                ;;
            *)
                if [ "$#" -gt 1 ]; then
                    true
                else
                    invalid "Invalid arguments provided."
                fi
                ;;
        esac
        ;;
esac

```

build/bin/.functions.sh
HariPalleti: Have added as a different function since these tools are installed only in the case OSCAP_ENABLED. Have removed sudo apt-get update since this was already done before.

build/bin/oscap-docker.sh
terc1997: It's because you're trying to translate the `if` logic straight to `switch/case`. You need to accomplish:
```
oscap-docker --oscap=<OSCAP_BINARY>] IMAGE_NAME OSCAP_ARGUMENT [OSCAP_ARGUMENT...]"
```
Instead of using positional arguments, use named ones:
```
oscap-docker  --oscap-bin <OSCAP_BINARY>  --image-name <IMAGE_NAME> --oscap-args [OSCAP_ARGUMENT...]
```
And then you can
```
case "$1" in
  -b|--oscap-bin)
    OSCAP_BINARY=$1
    shift
    ;;
  -i|--image-name)
    IMAGE_NAME=$1
    shift
    ;;
  -h|--help)
    usage
    exit 0
    ;;
  *)
     Invalid "Invalid arguments provided."
     exit 1
     ;;
esac
```
Something around these lines, this is just an snippet.

HariPalleti: Here the validation is mainly about the parameters passed to oscap-docker for this part of the command `oscap-docker [--oscap=<OSCAP_BINARY>]`. 
Which internally calls oscap command. oscap command has its own usage for the command. 
https://github.com/ibm-mas/cli/blob/harden/build/bin/oscap-docker.sh#L31
Don't think it is required to add the validation for the parameters passed to oscap command.

terc1997 (approved):

PR #1852, opened 2025-10-28
Sls app

Drashti-Malviya:
I’ve added the new parameters

1. ICN and
2. Subscription ID

and made updates to the secret path.

Please review and let me know if any further changes are required.

### DIFF
--- a/image/cli/mascli/functions/gitops_mas_config
+++ b/image/cli/mascli/functions/gitops_mas_config
@@ -333,7 +333,6 @@ function gitops_mas_config_noninteractive() {
   done
 
 
-
   [[ -z "$GITOPS_WORKING_DIR" ]] && gitops_mas_config_help "GITOPS_WORKING_DIR is not set"
   [[ -z "$ACCOUNT_ID" ]] && gitops_mas_config_help "ACCOUNT_ID is not set"
   [[ -z "$REGION_ID" ]] && gitops_mas_config_help "REGION_ID is not set"
@@ -343,7 +342,6 @@ function gitops_mas_config_noninteractive() {
   [[ -z "$ICN" ]] && gitops_mas_config_help "ICN is not set"
 
   
-
   
   [[ -z "$CONFIG_ACTION" ]] && gitops_mas_config_help "CONFIG_ACTION is not set"
   if ! [[ "$CONFIG_ACTION" =~ ^(upsert|remove)$ ]]; then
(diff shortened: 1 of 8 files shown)


image/cli/mascli/functions/gitops_mas_config
```diff
@@ -324,12 +333,15 @@ function gitops_mas_config_noninteractive() {
   done
 
 
-
   [[ -z "$GITOPS_WORKING_DIR" ]] && gitops_mas_config_help "GITOPS_WORKING_DIR is not set"
   [[ -z "$ACCOUNT_ID" ]] && gitops_mas_config_help "ACCOUNT_ID is not set"
   [[ -z "$REGION_ID" ]] && gitops_mas_config_help "REGION_ID is not set"
   [[ -z "$CLUSTER_ID" ]] && gitops_mas_config_help "CLUSTER_ID is not set"
   [[ -z "$MAS_INSTANCE_ID" ]] && gitops_mas_config_help "MAS_INSTANCE_ID is not set"
+  [[ -z "$SAAS_SUB_ID" ]] && gitops_mas_config_help "SAAS_SUB_ID is not set"
```
rbinns: SAAS_SUB_ID will only be present in cases where we are trying to configure MAS instance to point to a standalone SLS - This validation prevents the cases where we still use SLS provisioned with the instance.

image/cli/mascli/functions/gitops_mas_config
```diff
@@ -732,10 +744,12 @@ function gitops_mas_config() {
     fi
 
     if [ "${MAS_CONFIG_TYPE}" == "sls" ]; then
```
rbinns: This case will only support cases where we use stand alone sls - it does not support the case where SLS provisioned with MAS CORE is used.

tekton/src/tasks/gitops/gitops-suite-config.yml.j2
```diff
@@ -47,6 +47,10 @@ spec:
     - name: smtp_config_ca_certificate_file
       type: string
       default: ""
+    - name: ibm_customer_number
```
rbinns: given that these will not always be supplied, we need default values for these parameters so that we don't force them to be specified.

rbinns (changes requested):

image/cli/mascli/functions/gitops_mas_config
Drashti-Malviya: i have update this ,please check.

image/cli/mascli/functions/gitops_mas_config
Drashti-Malviya: i have update this ,please check.

tekton/src/tasks/gitops/gitops-suite-config.yml.j2
Drashti-Malviya: i have update this ,please check.

rbinns (dismissed):

whitfiea (changes requested):
PR title needs to be better than just `Sls app`

Drashti-Malviya:
revert this changes.

PR #1872, opened 2025-11-05
[patch] Add to AI Service support for watsonX wx_verify

karol-czarnecki:
As a developer I need to add variable wx_verify to secret - and set to true once customer wants to use self sign certificate.

https://jsw.ibm.com/browse/MASAIB-1514


whitfiea (changes requested):
A description is needed in this PR

whitfiea (approved):

PR #1874, opened 2025-11-06
[patch] Support setting routing mode during suite install

stonepd:
### PRE-RELEASE
First Merge:
https://github.com/ibm-mas/python-devops/pull/133
https://github.com/ibm-mas/ansible-devops/pull/1974

### Description
Set the mas_routing_mode parameter when running the CLI mas install  command. Include this parameter in the install pipeline run.

### Related Issues
https://jsw.ibm.com/browse/MASCORE-8699

### Testing
Tested by running CLI installations against the pfvtpds personal FVT cluster


python/src/mas/cli/install/argParser.py
```diff
@@ -159,6 +159,13 @@ def isValidFile(parser, arg) -> str:
     action="store_const",
     const="false"
 )
+masAdvancedArgGroup.add_argument(
+    "--routing",
+    dest="mas_routing_mode",
+    required=False,
+    help="Configure MAS with a flat or subdomain routing",
```
IanBoden: it's a little confusing using "flat" and then the option being "path"

python/src/mas/cli/install/app.py
```diff
@@ -469,7 +470,21 @@ def configOperationMode(self):
...
         ])
-        self.operationalMode = self.promptForInt("Operational Mode", default=1)
+        self.operationalMode = self.promptForInt("Operational Mode", default=1, min=1, max=2)
+
+    @logMethodCall
+    def configRoutingMode(self):
+        if self.showAdvancedOptions:
+            self.printH1("Configure Routing Mode")
+            self.printDescription([
+                "Maximo Application Suite can be installed so it can be accessed with single domain URLs (path mode) or multi-domain URLs (subdomain mode):",
```
IanBoden, line 480: I think we should only show if in FC I don't know if this is enough:
`if isVersionEqualOrAfter('9.2.0', self.getParam("mas_channel")):`
or if you need:
`if isVersionEqualOrAfter('9.2.0', self.getParam("mas_channel")) or self.getParam("mas_channel") == '9.2.x-feature':`
I don't know if we also want the description to say say "(feature currently in development)". And we should probably have the default as subdomain for now.

IanBoden (changes requested):

python/src/mas/cli/install/argParser.py
stonepd: what do you want it to say?

python/src/mas/cli/install/app.py
stonepd: Now only given this option for 9.2.0 installs, excluding 9.2.0 feature channel

python/src/mas/cli/install/argParser.py
stonepd: I changed `flat` to `path`

IanBoden (approved):

PR #1876, opened 2025-11-06
[minor] Add support for configuring EFS CSI driver via ArgoCD

tomklapiscak:


tomklapiscak:
Branch name breaking docker build (cannot use underscores), closing this PR and reraising under a different branch name.

PR #1888, opened 2025-11-11
[doc] Clarifying RHOCP support statement

gbanfo:
Cleaning up the RHOCP support statement to make it consistent with the overview doc on our EUS Term 1 subscription support


whitfiea (changes requested):
Are we expecting more changes like this? Can you combine them into one PR rather than having a separate PR per file

whitfiea (approved):

PR #1915, opened 2025-11-24
[minor] Integrate aiservice into mas-install pipeline

Bhautik-Vala:
Enable mas-install pipeline to install AI Service as well - eventually combining aiservice-install pipeline into mas-install pipeline -> removing aiservice-install pipeline.

Story Link: https://jsw.ibm.com/browse/MASR-6486
Test story: https://jsw.ibm.com/browse/MASR-6487 
Related PR:
https://github.com/ibm-mas/python-devops/pull/141

Testing:
Screenshot 2025-11-25 at 2 23 15 PM
Screenshot 2025-11-25 at 3 04 54 PM


durera (commented):
At the moment we've basically attached the aiservice pipeline to the side of the existing MAS one, creating duplication in places ... what I'm looking for here is full integration into the pipeline 

Image

- sls - we are already installing sls in the pipeline, we shouldn't need a seperate stage, if customer is only installing aiservice, installing mas+aiservice, or just installing mas on it's own then the existing sls stage in the pipeline should be only place we "do sls stuff"
- dro - as above, dro is a cluster scoped install, we don't have a mas dro and an aiservice dro
- mongodb - same, our strategy is one mongodbcluster per ocp cluster, and everything that uses it uses it

I'd expect something like this to be what we are inserting:

- minio (e.g. run after ibm-catalogs)
- db2-aiservice (e.g. run after cert-manager)
- odh (e.g. run after ibm-catalogs)
- aiservice (run after existing sls, existing dro, db2-aiservice, odh)

Also, let's squash aiservice-tenant into a second step in aiservice task, see how I have done this for all mas apps, as we are trying to simplify the pipeline/make it easier for customers to read/understand/process.

Bhautik-Vala:
@durera 
Regarding separation of two similar tasks (e.g., sls/dro/mongo): when I tried to integrate the AIServices pipeline into MAS, the main issue was that when standalone AIServices is installed, we have to pass aiservice_instance_id as mas_instance_id to the SLS role. I thought a lot about how we could use a single task for both scenarios (i.e., standalone AIServices and with MAS). We want the dynamic nature of mas_instance_id’s value, but this can’t be achieved through Jinja templating because at runtime only we know whether mas_instance_id should be mas_instance_id or aiservice_instance_id for the SLS/DRO/Mongo role.

So I kept two tasks, but only one gets executed depending on the scenario.

When we install MAS + AIServices, this is the list of tasks:

Screenshot 2025-12-10 at 10 56 39 AM

Can you please suggest any tips on how we can dynamically set the value of mas_instance_id within a single task depending on the scenario?

Also, I will squash aiservice-tenant into the aiservice task.

Bhautik-Vala:
@durera I have Integrate aiservice-tenant into the aiservice task.
Screenshot 2025-12-10 at 1 08 40 PM

durera:
> Can you please suggest any tips on how we can dynamically set the value of mas_instance_id within a single task depending on the scenario?

Assuming we are using mas_instance_id here to generate a "fake" slsfcfg, mongocfg, etc because that's just what the code in the aiservice provision expects then I suggest something like this (e.g. in the sls task):


```
  params:
    - name: mas_instance_id
      type: string
      description: Instance ID
      default: "" # By default, no config will be generated
    - name: aiservice_instance_id
      type: string
      description: AIService Instance ID - when installing AI Service in standalone mode we still need to generate the MAS slscfg object, because it is used to populate the AIService configuration
      default: ""

  stepTemplate:
    env:
      - name: MAS_INSTANCE_ID
        value: $(params.mas_instance_id)
      - name: AISERVICE_INSTANCE_ID
        value: $(params.aiservice_instance_id)

  steps:
    - name: sls
      script: |
        #!/bin/bash
        
        # Support AI Service standalone
        if [[ "$MAS_INSTANCE_ID" == "" ]] && [[ "$AISERVICE_INSTANCE_ID" != "" ]]; then
            echo "Setting MAS_INSTANCE_ID based on AISERVICE_INSTANCE_ID for standalone AI Service install"
            MAS_INSTANCE_ID=AISERVICE_INSTANCE_ID
        fi
        /opt/app-root/src/run-role.sh sls
```

This will make the task use aiservice-instance-id as a "fake" mas-instance-id when the role is ran, so that the MAS config resource file is still generated and can be used later in the pipeline.

Bhautik-Vala:
Hi @durera 
have done the changes regarding of sls/dro/mongo and tested for mas install and mas aiservice-install both here are the results:
Screenshot 2025-12-11 at 3 04 18 PM
Screenshot 2025-12-11 at 11 51 00 AM
can you please review again. Thank you.

tekton/src/pipelines/mas-install.yml.j2
```diff
@@ -63,6 +65,9 @@ spec:
         - input: "$(params.grafana_action)"
           operator: in
           values: ["install"]
+        - input: "$(params.standalone_aiservice_install)"
```
durera: Looks great after the updates to integrate into the mainline trunk of the pipeline.  I think the last thing now is to adjust the behaviour "switch" logic so that the overall pipeline structure is all in place ready for the three scenarios we want the pipeline to support (even if the cli won't support them all initially for now) ...

Currently we have a boolean flag `standalone_aiservice_install` ... but what we need is logic to be based around `mas_instance_id == ""` and `aiservice_instance_id == ""`

So, e.g. the AI Service tasks when clauses should be gated on the logic:
```
        - input: "$(params.aiservice_instance_id)"
          operator: notin
          values: [""]
```

e.g. and the MAS tasks when clauses should be gated on the logic:
```
        - input: "$(params.mas_instance_id)"
          operator: notin
          values: [""]
```

The control whether the MAS and AI Service parts of the pipeline run, should be based entirely on whether the instance IDs are set or not .... e.g. I should be able to run four primary scenarios:

- mas_instance_id == "" aiservice_instance_id == "" ... not too worried about this one .. it might be useful in the future, but for now we can effectively treat this as user error and not really care what the pipeline does in this case
- mas_instance_id == "" aiservice_instance_id == "myaiservice" ... ai service standalone install (would be in namespace `aiservice-myaiservice-pipelines`)
- mas_instance_id == "mymas" aiservice_instance_id == ""   ... mas without aiservice (would be in namespace `mas-mymas-pipelines`)
- mas_instance_id == "mymas" aiservice_instance_id == "myaiservice"  ... mas with aiservice (would be in namespace `mas-mymas-pipelines`)


We also need to consider the different SLS scenarios supported, and make sure AI Service is able to cater to them:
- If we install MAS + AI Service together, they will share an SLS and a pool of apppoints ... for this to work the AI Service integration with SLS has to be at the aiservice-tenant level, not the aiservice level .. otherwise we can't have a tenant per MAS, we need an entire aiservice install per MAS, which means the tenant is kind of redundant concept.  Can you confirm AI Service -> SLS is at the tenant level?

Bhautik-Vala: Yes will change the conditions that based on instance id's rather than flag and regarding SLS thing - AI Service integration with SLS is at **aiservice-tenant** level as AIServiceApp CR does not contain info related to SLS but AIServiceTenant CR contains all info related to SLS.

durera (commented):
This is ready to merge ... approval pending the readiness of related PRs in python-devops and ansible-devops

Bhautik-Vala:
> This is ready to merge ... approval pending the readiness of related PRs in python-devops and ansible-devops

@durera For phase-1 we need only python-devops changes in this cli PR, right now python-devops is ready with required changes and I've tested that - can you please merge python-devops and this cli PR.

durera (approved):

PR #1927, opened 2025-11-27
[minor] Add aiservice mirror airgap

IanBoden:
creates a new command for mirroring ai service moving code from the existing mirror command into a common library for both to use.

can't merge till python-devops and ansible-devops have been updated and mas_devops.tar.gz is removed


durera:
Closing PR due to old age, if this is still needed please re-open the PR and organize a review via Slack

PR #1939, opened 2025-12-01
[patch] disable slackclient

terc1997:
## Changes
Temporarily disable the Slack notifications until the new changes from `slack-sdk` gets propagated into CLI.


terc1997:
Cancelled.

PR #1966, opened 2025-12-16
[patch] Fix October catalog metadata for s390x and ppc64le

prveenkumr:


durera (commented):
I'll fix it for you this time, but you need to install `pre-commit` so that linting etc is handled client side before you push code .. this is why your build is failing, and it can be easily caught before raising a PR.

durera (dismissed):

durera (approved):

Issue #1812, opened 2025-10-02
Include mas_cluster_issuer in interactive `mas install`


mudspringhiker:
### Feature and motivation

A customer wants to have interactive mas install prompt for the clusterissuer yaml file. Currently, this can only be done non-interactively. For example:

`export IBM_ENTITLEMENT_KEY=x
mas install --mas-catalog-version v9-250403-amd64 --ibm-entitlement-key $IBM_ENTITLEMENT_KEY \
  --mas-channel 9.0.x --mas-instance-id certtest2 --mas-workspace-id certtest2 --mas-workspace-name "certtest2" \
  --non-prod \
  --storage-class-rwo "rook-ceph-block" --storage-class-rwx "rook-cephfs" \
  --storage-pipeline "rook-cephfs" --storage-accessmode "ReadWriteMany" \
  --uds-email "xxx@mas.com" --uds-firstname "myname" --uds-lastname "lastname" \
  --mongodb-namespace "mongoce" \
  --mas_cluster_issuer mas-ca-issuer
  --accept-license --no-confirm
`

### Usage example

Checked whether this can be done interactively but can't see it.

<img width="437" height="163" alt="Image" src="https://github.com/user-attachments/assets/3029e6d2-e643-492e-ab8b-f08141460c29" />

mudspringhiker:
ref: TS020066306

terc1997:
Hi, @durera @mudspringhiker. I'm gonna deliver on the current sprint.

PR #1983, opened 2025-12-26
[minor] Support BYO ClusterIssuer

terc1997:
### Related issues
- https://github.com/ibm-mas/cli/issues/1812

### Changes
- Implement override for `mas_cluster_issuer` parameter. 

#### Override without provider
Pasted Graphic
Pasted Graphic 1

#### Override with provider - CIS
Pasted Graphic 3
Pasted Graphic 4

#### No override with no provider
Pasted Graphic 5
Pasted Graphic 6

#### No override with a provider - CIS
Pasted Graphic 7
Pasted Graphic 8


python/src/mas/cli/install/app.py
durera, line 1: Why do we want to override the cluster issues when we are setting up the DNS integration?  It means that we won't be deploying with the clusterissuer that we are setting up for the install?

terc1997: From the issue raised by the customer, I understood that they will provide a custom cluster issue and reference in the mas install. If you see the non-interactive method that's what they're doing.

durera: it should be exclusive ...

e.g.

if you are setting up dns integration then you don't get to pick the cluster issuer name, because for the setup to work you have to be using the one that we are creating that will work with the dns integration

if you are not setting up dns integration then we should support the option to supply the name of a pre-created cluster issuer.

I don't think it's a valid usecase to ask the installer to manage DNS via one of the supported providers and also configure it to use a custom issuer

e.g.

```python
            configureDomainAndCertMgmt = self.yesOrNo('Configure domain & certificate management')
            if configureDomainAndCertMgmt:
                # do stuff for CIS, R5, CF etc.  We will always use the clusterissuer that we set up tied to the DNS management we are doing
            else:
                # do you want MAS to set up it's own (self-signed) clusterissuer, or do you want to tell it to use one that exists - ideally pull up a list of clusterissuers on the cluster and ask them to pick which one they want to use rather than asking them to just type it out

```

terc1997: Ok. Gotcha ! I'll make the changes.

python/src/mas/cli/install/app.py
```diff
@@ -886,13 +887,34 @@ def configDNSAndCerts(self):
                     # Use MAS default self-signed cluster issuer with the default domain
                     self.setParam("dns_provider", "")
                     self.setParam("mas_domain", "")
-                    self.setParam("mas_cluster_issuer", "")
```
durera, line 900: What's the reason we don't need this set to empty here now, wouldn't we still want this set to "" in this case?

python/src/mas/cli/install/app.py
```diff
@@ -886,13 +887,34 @@ def configDNSAndCerts(self):
...
                 self.manualCerts = self.yesOrNo("Configure manual certificates")
                 self.setParam("mas_manual_cert_mgmt", self.manualCerts)
                 if self.getParam("mas_manual_cert_mgmt"):
                     self.manualCertsDir = self.promptForDir("Enter the path containing the manual certificates", mustExist=True)
                 else:
                     self.manualCertsDir = None
+            else:
+                # Configuring domain
+                if self.yesOrNo('Configure custom domain'):
```
durera, line 909: I find the current flow a bit confusing ...

Do you want to configure domain and cert management? No
Do you want to configure custom domain? Yes

Feel like most customers would have answered Yes to the first one, even if they wanted what's in the second path.

Think we should have a single path through this taking customers through the options, with a single prompt asking them to pick how they want to manage the certs:

```
if (Do you want to configure domain and cert management?)
    Configure custom domain

    Choose certificate management mode
    1. Certificate-Manager with DNS integration
    2. Certificate-Manager with existing ClusterIssuer
    3. Certificate-Manager with auto-generated self-signed certificate
    4. Manual certificate management

    > 
``` 

This can then be easily extended by @rawa-resul for his work to add support for Issuer as an alternative to ClusterIssuer.

durera (changes requested):

durera:
Closing due to old age.  If still working on this raise a new PR when it's ready for review/merge

PR #1989, opened 2026-01-02
[patch] Update installation of ArcGIS on CLI : MASCORE-10767

Alfishan:
### Summary:
Refactored ArcGIS dependency configuration to support both Manage and MREF applications with proper installation ordering.

- Moved ArcGIS settings to generic location (app.py) from Manage-specific settings for reusability across Manage and MREF
- Updated installation order in Tekton pipeline to ensure ArcGIS installs after both Manage and Facilities apps (dependencies)
- Refactored configuration template (arcgis.yml.j2) to use generic settings structure

### Tested Using cli:18.5.1-pre.alfs-masr-6128-arcgis image. 
image
image

### DIFF
--- a/python/src/mas/cli/install/app.py
+++ b/python/src/mas/cli/install/app.py
@@ -372,6 +372,61 @@ class InstallApp(BaseApp, InstallSettingsMixin, InstallSummarizerMixin, ConfigGe
                 else:
                     self.promptForString("Install namespace", "grafana_v5_namespace", default="grafana5")
                     self.promptForString("Grafana storage size", "grafana_instance_storage_size", default="10Gi")
+    @logMethodCall
+    def arcgisSettings(self) -> None:
+        """
+        Configure ArcGIS as a shared dependency for Manage and Facilities.
+        This method detects if either Manage (with Spatial) or Facilities is selected
+        and prompts for ArcGIS installation accordingly.
+        """
+        needsArcGIS = False
+        apps_requiring_arcgis = []
+        
+        # Check if Manage with Spatial component is selected
+        if self.installManage and "spatial=" in self.getParam("mas_appws_components"):
+            needsArcGIS = True
+            apps_requiring_arcgis.append("Maximo Manage (Spatial)")
+        
+        # Check if Facilities is selected (MAS 9.1+)
+        if self.installFacilities:
+            needsArcGIS = True
+            apps_requiring_arcgis.append("Maximo Real Estate and Facilities")
+        
+        # Only prompt if ArcGIS is needed and we're on MAS 9.x
+        # Check the appropriate channel based on what's being installed
+        if needsArcGIS:
+            channel = self.getParam("mas_app_channel_manage") or self.getParam("mas_app_channel_facilities")
+            if channel and channel.startswith("9."):
+                # Build description based on what's being installed
+                description = [
+                    "",
+                    "Geospatial capabilities require a map server provider",
+                    "You may choose your preferred map provider later or you can enable IBM Maximo Location Services for Esri now"
+                ]
+                
+                # Add specific details based on installed applications
+                if len(apps_requiring_arcgis) == 1:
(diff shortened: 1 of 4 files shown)


python/src/mas/cli/install/app.py
```diff
@@ -373,6 +373,61 @@ def configGrafana(self) -> None:
...
+        # Check if Facilities is selected (MAS 9.1+)
+        if self.installFacilities:
+            needsArcGIS = True
+            apps_requiring_arcgis.append("Maximo Real Estate and Facilities")
+
+        # Only prompt if ArcGIS is needed and we're on MAS 9.x
+        # Check the appropriate channel based on what's being installed
+        if needsArcGIS:
+            channel = self.getParam("mas_app_channel_manage") or self.getParam("mas_app_channel_facilities")
+            if channel and channel.startswith("9."):
```
terc1997: In python-devops there is a function to verify versions. Please refer to [isVersionEqualOrAfter](https://github.com/ibm-mas/python-devops/blob/0ab0147871f84226c4e5a2a68b3140f70f3c8e22/src/mas/devops/utils.py#L45-L76) and make the changes.

python/src/mas/cli/install/app.py
```diff
@@ -1260,6 +1316,14 @@ def nonInteractiveMode(self) -> None:
                 if not vars(self.args).get("aiservice_instance_id") and value is not None and value != "":
                     self.setParam("manage_bind_aiservice_tenant_id", value)
 
+            # ArcGIS settings
+            elif key == "install_arcgis":
+                if value is not None and value != "":
```
terc1997, line 1321: For non interactive mode we need to add a special handling for the arcgis parameter, otherwise the customer will be able to add the ArcGIS without the Facilities or Spatial. After parsing all parameters, the code needs to check if alongside with the ArcGIS, we have either Facilities or Spatial in the parameters.

python/src/mas/cli/install/app.py
```diff
@@ -373,6 +373,61 @@ def configGrafana(self) -> None:
...
+
+                # Add specific details based on installed applications
+                if len(apps_requiring_arcgis) == 1:
+                    description.append(f"This includes ArcGIS Enterprise as part of {apps_requiring_arcgis[0]} (Additional AppPoints required).")
+                else:
+                    description.append(f"This includes ArcGIS Enterprise for {' and '.join(apps_requiring_arcgis)} (Additional AppPoints required).")
+
+                self.printDescription(description)
+
+                if self.yesOrNo("Include IBM Maximo Location Services for Esri"):
```
terc1997, line 416: Spatial can be installed without ArcGIS ? If no, the code should have two flows:
1. Blocking the installation in case of ArcGIS is missing for Spatial.
2. Allowing the installation of Facilities when ArcGIS is not installed, since it's not required.

terc1997 (changes requested):

python/src/mas/cli/install/app.py
Alfishan: I have replaced it with [isVersionEqualOrAfter](https://github.com/ibm-mas/python-devops/blob/0ab0147871f84226c4e5a2a68b3140f70f3c8e22/src/mas/devops/utils.py#L45-L76).

python/src/mas/cli/install/app.py
Alfishan: Made changes [here](https://github.com/ibm-mas/cli/pull/1989/files#diff-6ad4c0f0e7c06f1af6575fb9a1b69a9093ca48c9d5fd0da42ab522e92512189eR1494) for non interactive mode.

python/src/mas/cli/install/app.py
Alfishan: Done. Added this for now. As Thiago is on vacation and I can not confirm it.

Alfishan: @terc1997 Spatial can be installed without ArcGIS as per Thiago. I will be doing this changes. 
image

terc1997 (approved):
LGTM

PR #1992, opened 2026-01-05
[patch] Generate pipeline artifacts

unnati-solanki-git:
https://jsw.ibm.com/browse/MASR-6769

Locally tested pipeline and tasks:
image

image

### DIFF
--- a/tekton/src/pipelines/mas-fvt-deprovision-after.yml.j2
+++ b/tekton/src/pipelines/mas-fvt-deprovision-after.yml.j2
@@ -142,7 +142,7 @@ spec:
           operator: in
           values: ["uninstall"]
 
-    # 2. Delete the named cluster
+    # 3. Delete the named cluster
     # -------------------------------------------------------------------------
     - name: deprovision-cluster
       taskRef:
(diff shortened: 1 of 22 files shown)


tekton/src/pipelines/mas-fvt-deprovision-after.yml.j2
```diff
@@ -132,3 +151,25 @@ spec:
       runAfter:
         - cos-deprovision
         - kafka-deprovision
+
+  # Finally
```
durera: We will have to break with the standard approach here ... the last task in the pipeline deleted the cluster, so the finally will never work :D 

I think we should still try to record the pipeline-finish, but I think we need to move it into the main pipeline as a special case, and run it before the deprovision-cluster step

unnati-solanki-git: You are right. This will not work as we already lost the cluster by the time it reaches to finally block. 

I've moved it to mas-fvt-finally.yml.j2 for this special case.

unnati-solanki-git: Considering the limitation of pipeline_status param, combined deprovision and pipeline-finish for finally pipeline (special case)

image

durera (approved):

PR #2033, opened 2026-02-03
Update cli docs and add in cleanup workspaces task

whitfiea:
Dev to Dev PR only.

Adds a `mas backup` command doc and a `backup and restore` guide doc.

Also adds a clean-workspaces task that runs in the finally block of the backup pipeline. This will remove data from the /workspace/config and /workspace/backup folders to that when the pipeline is re-run (i.e. another backup later on) then the same backup-pvc will be clean of any data and not fill up over time.


docs/guides/backup-restore.md
```diff
@@ -0,0 +1,543 @@
...
+
+The backup creates a compressed archive that can be stored locally or uploaded to cloud storage (S3 or Artifactory).
+
+### Ansible DevOps Integration
+
+The `mas backup` command launches a Tekton pipeline that executes the following Ansible roles from the [IBM MAS DevOps Collection](https://ibm-mas.github.io/ansible-devops/):
+
+- [`ibm.mas_devops.ibm_catalogs`](https://ibm-mas.github.io/ansible-devops/roles/ibm_catalogs/) - Backs up IBM Operator Catalog definitions
+- [`ibm.mas_devops.cert_manager`](https://ibm-mas.github.io/ansible-devops/roles/cert_manager/) - Backs up Certificate Manager configurations
+- [`ibm.mas_devops.mongodb`](https://ibm-mas.github.io/ansible-devops/roles/mongodb/) - Backs up MongoDB Community Edition databases
```
sanjayprab: both Mongo instance and Mongo database

docs/guides/backup-restore.md
```diff
@@ -0,0 +1,543 @@
...
+
+!!! warning
+    IBM Cloud Databases for MongoDB and other external MongoDB providers are not supported by the backup process. You must use their native backup mechanisms.
+
+### Certificate Manager
+
+Specify the certificate manager provider used in your environment:
+
+- **Red Hat Certificate Manager** (`--cert-manager-provider redhat`) - Default option
+- **IBM Certificate Manager** (`--cert-manager-provider ibm`) - For IBM Cloud Pak deployments
```
sanjayprab: do we support IBM cert manager??

docs/guides/backup-restore.md
```diff
@@ -0,0 +1,543 @@
...
+Storage Requirements
+-------------------------------------------------------------------------------
+
+### Backup Storage Sizing
+
+The backup storage size depends on several factors:
+
+| Component | Typical Size | Notes |
+|-----------|-------------|-------|
+| MAS Configuration | < 1 GB | Core MAS custom resources and configurations |
```
sanjayprab: < 1 GB seems to be too big a scale for this, should we mention this in MB? 
we can use GB for databases like mongodb and db2

docs/guides/backup-restore.md
```diff
@@ -0,0 +1,543 @@
...
+- Verify successful completion
+
+### Backup Artifacts
+
+Backups are stored in the pipeline namespace PVC at:
+
+- **Backup Directory**: `/workspace/backups`
+- **Config Directory**: `/workspace/configs`
+
+The final backup archive is named: `mas-backup-{instance-id}-{backup-version}.tar.gz`
```
sanjayprab: i don't think we are including {instance-id} in the archive filename

docs/guides/backup-restore.md
whitfiea: added the instance part now

docs/guides/backup-restore.md
whitfiea: no, I removed it from the docs now

docs/guides/backup-restore.md
whitfiea: I have altered the sizings based on the files in the test. I also removed the recommended sizings as that was based on MAS workspaces which is not relevant

docs/guides/backup-restore.md
whitfiea: ah yes missed that. changed now

sanjayprab (approved):

PR #2036, opened 2026-02-05
[minor] New mirror capability based on ISC

durera:
## New Mirror Function

`mas mirror`

This does not immediately replace the `mas mirror-images` command, but will be made available alongside it as a tech preview in effect.

However, eventually it will replace the old `mirror-images` command.

- Improved UI with better progress reporting
- Clearer logging
- Utilizes `oc-mirror` with `ImageSetConfiguration` resources instead of `oc image mirror`
- Removed dependency on `ibm-pak` and IBM CASE metadata for image mirroring, we use https://github.com/ibm-mas/image-set-configs as the source of truth for what images need to be mirrored

image

## Install Function Updates

Also, fixed a number of issues in `install` while testing these changes:

- Add `NoReturn` type hint for `fatalError()` method
- Fix numerous `bool` types added to `self.params` (which should be all strings)
- Fix generateJDBCCfg() type hint in `Db2SettingsMixin`
- Fix `dynamicClient` type hint in `ManageSettingsMixin`
- Fix `generateMongoCfg` type hint in `MongoDbSettingsMixin`
- Disabled OCP version check when `--dev-mode` is set (because we don't load the catalog metadata files containing this information)
- Added various asserts to check for `None` when variables are expected to be defined before they are used in scenarios where a non-None value is expected
- Fixed code that converts `cos_type` `odf` to `ocs` internally, because we still use the old name in the ansible code
- Added type hint for `self.approvals`
- Improved logging in the `displayMixin`

## Testing
Successful airgapped install in Fyre:

image


sanjayprab:
looks good!!

PR #2044, opened 2026-02-06
[minor] Support optional ingress controller config for PBR

dixitgsathwara:
## Description

- patching ingress controller for the path based routing during the cli instead of suite install
- add test cases for the routing mode functionality 

## Testing
image

image


python/src/mas/cli/install/app.py
durera: I think at this point we should explain the requirements

> **Routing Configuration**
> Maximo Application Suite can be configured in one of two ways:
> - Single domain with path-based routing across the suite - https://{yourdomain)/admin
> - Single domain with subdomain-based routing across the suite - https://admin.{yourdomain}
> 
> Path-based routing requires a specific OpenShift ingress controller configuration, for more information refer to:
> https://somelinktotheocpdocsaboutthistopic
> 

We should know enough to substitute in the actual domain the user will be using at this point I think.


If the cluster doesn't support PBR, AND the user has the necessary permissions, show this additional message:
> Your cluster ingress currently does not support path-based routing
> 
> Configure ingress namespace ownership policy to enable path-based routing for MAS? [y/n]
> 
If they choose y and this fails we should handle it as a fatalError and exit

If the cluster doesn't support PBR AND the user DOES NOT have the necessary permissions, show something like this:
> Your cluster ingress currently does not support path-based routing, if you wish to configure MAS with path-based routing contact your OpenShift administrator

Then show them the appropriate options only if PBR is supported after this.  If the ingress still doesn't support PBR at this stage then we don't need to show them a prompt and just state something like:

> MAS will be configured to use subdomain-based routing

Also, as we are adding new RBAC needs to the cli to get/set ingress controller, please include an update to: https://github.com/ibm-mas/cli/tree/master/rbac/install/user to reflect this, otherwise the minimal RBAC install will be broken.

durera (changes requested):

python/src/mas/cli/install/app.py
dixitgsathwara: Done!

durera (dismissed):

durera:
Reviewed and approved, but note that this should only be merged **after** https://github.com/ibm-mas/ansible-devops/pull/2105#pullrequestreview-3785800225 has been merged

dixitgsathwara:
> Reviewed and approved, but note that this should only be merged **after** [ibm-mas/ansible-devops#2105 (review)](https://github.com/ibm-mas/ansible-devops/pull/2105#pullrequestreview-3785800225) has been merged

https://github.com/ibm-mas/python-devops/pull/183 this pr is also required to be merged first

durera (approved):

PR #2055, opened 2026-02-10
[patch] Fix NoSuchCatalogError handling for AI Service

Bhautik-Vala:
Handling NoSuchCatalogError Exception in dev-mode ( non-interactive mode ) for **mas aiservice-install** 
Tested this changes by running mas aiservice-install in dev-mode and without dev-mode.

### DIFF
--- a/python/src/mas/cli/aiservice/install/app.py
+++ b/python/src/mas/cli/aiservice/install/app.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 # *****************************************************************************
-# Copyright (c) 2024, 2025 IBM Corporation and other Contributors.
+# Copyright (c) 2024, 2026 IBM Corporation and other Contributors.
 #
 # All rights reserved. This program and the accompanying materials
 # are made available under the terms of the Eclipse Public License v1.0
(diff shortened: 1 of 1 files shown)


python/src/mas/cli/aiservice/install/app.py
```diff
@@ -347,7 +347,12 @@ def nonInteractiveMode(self) -> None:
                 self.fatalError(f"Unknown option: {key} {value}")
 
         # Load the catalog information
-        self.chosenCatalog = getCatalog(self.getParam("mas_catalog_version"))
+        try:
+            self.chosenCatalog = getCatalog(self.getParam("mas_catalog_version"))
+            catalogSummary = self.processCatalogChoice()
+            self.printDescription(catalogSummary)
+        except NoSuchCatalogError:
+            pass
```
whitfiea, line 355: if we are passing then chosenCatalog would not be set and so cause errors later on. Also getCatalog is also called later on in this file but has not try block

whitfiea (changes requested):

python/src/mas/cli/aiservice/install/app.py
Bhautik-Vala: chosenCatalog is only getting used in processCatalogChoice() function, which we call only if we are able to find catalog metadata from python-devops.
And about the second getCatalog — this part only runs if we are not in dev-mode. So I think if we are not in dev mode, then the catalog should be the static one, and it should be present in the python-devops metadata. If not, then showing an error as NoSuchCatalog makes sense.

Bhautik-Vala: and the second occurence of getCatalog is for interactive mode.

whitfiea (approved):

PR #2061, opened 2026-02-12
Add Manage Db2 and Application to MAS restore pipelines

sanjayprab:
DEV->DEV PR FOR REVIEW

Adding manage db2 and app restore to the MAS restore pipelines.
Documentation will be covered in a different PR

depends on 
python-devops - https://github.com/ibm-mas/python-devops/commit/e8b12137d0d24dbfab8f2ff17a01e22adb1df6ac
ansible-devops - https://github.com/ibm-mas/ansible-devops/pull/2118

TESTS:

MAS PIPELINE WITH MANAGE DB2 AND APP RESTORE - FINISHED SUCCESSFULLY

Screenshot 2026-02-12 at 00 54 18

Manage Foundation successfully restored

Screenshot 2026-02-12 at 00 52 12

Can see users for example sanjaytest1 restored from backup
Screenshot 2026-02-12 at 00 52 32


python/src/mas/cli/restore/app.py
```diff
@@ -137,12 +146,15 @@ def restore(self, argv):
             if self.args.mas_domain_on_restore is None:
                 self.promptForMASConfiguration()
 
-            if self.args.include_slscfg_from_backup is None:
+            if self.getParam("include_sls") == "true":
                 self.promptForSLSConfiguration()
 
-            if self.args.include_drocfg_from_backup is None:
+            if self.getParam("include_dro") == "true":
```
whitfiea, line 152: what is the logic in this change?

whitfiea (changes requested):

python/src/mas/cli/restore/app.py
sanjayprab: Two reasons,
1. I want to show this only when user opts to include dro/sls in the above options.
2. Also, self.args.include_drocfg_from_backup or self.args.include_slscfg_from_backup is not None even when we don't pass anything. so this wasnt displayed.

whitfiea (approved):

PR #2066, opened 2026-02-13
[minor] Added changes to generate EFS CSI Operator template for gitops

dpdifferent1:
Added the task gitops-efs-csi-driver in gitops-cluster pipeline that calls the  function gitops-efs-csi-driver-operator for merging EFS CSI Driver related params into gitops envs. The EFS CSI Driver app is getting created on ArgoCD.
Screenshot 2026-02-13 at 12 24 32 PM
Screenshot 2026-02-13 at 12 24 52 PM
[https://jsw.ibm.com/browse/MASR-6337](url)

Added changes for jdbc_instance_name as part of integration testing


whitfiea (changes requested):
What is the diffference between this and the existing function https://github.com/ibm-mas/cli/blob/fb2db6f738c613f05320471d4cf8b309f4524f86/image/cli/mascli/functions/gitops_efs_csi_driver ?

dpdifferent1:
> What is the diffference between this and the existing function https://github.com/ibm-mas/cli/blob/fb2db6f738c613f05320471d4cf8b309f4524f86/image/cli/mascli/functions/gitops_efs_csi_driver ?

The existing function had been created for govcloud implementation hence created a new function

dpdifferent1:
Removed the new function & clubbed necessary changes into the older function gitops-efs-csi-driver

image/cli/mascli/templates/gitops/appset-configs/cluster/instance/configs/ibm-mas-jdbc-config.yaml.j2
```diff
@@ -8,7 +8,7 @@ mas_config_api_version: "config.mas.ibm.com"
 use_postdelete_hooks: {{ USE_POSTDELETE_HOOKS }}
 
 jdbc_type: {{ JDBC_TYPE }}
-jdbc_instance_name: <path:{{ SECRETS_PATH }}:{{ SECRET_KEY_JDBC_INSTANCE_NAME }}>
+jdbc_instance_name: {{ SECRET_KEY_JDBC_INSTANCE_NAME }}
```
whitfiea: why was this changed? from the function the SECRET_KEY_JDBC_INSTANCE_NAME would just be acountid/clusterid/.. etc with no <path

image/cli/mascli/functions/gitops_efs_csi_driver
```diff
@@ -51,10 +55,21 @@ function gitops_efs_csi_driver_noninteractive() {
...
+        ;;
+      --efs-catalog-source-namespace)
+        export EFS_CATALOG_SOURCE_NAMESPACE=$1 && shift
+        ;;
+      --efs-channel)
+        export EFS_CHANNEL=$1 && shift
+        ;;
+      --efs-subscription-source-namespace)
+        export EFS_SUBSCRIPTION_SOURCE_NAMESPACE=$1 && shift
+        ;;
```
whitfiea, line 72: we need some defaults set for these as this would break any using this function. the defaults should be what was set previously in the template

image/cli/mascli/functions/gitops_cluster
```diff
@@ -370,6 +374,7 @@ function gitops_cluster_noninteractive() {
   [[ -z "$CLUSTER_ID" ]] && gitops_cluster_help "CLUSTER_ID is not set"
   [[ -z "$REGION_ID" && -z "$SM_AWS_REGION" ]] && gitops_cluster_help "REGION_ID or SM_AWS_REGION is not set"
   [[ -z "$CLUSTER_URL" ]] && gitops_cluster_help "CLUSTER_URL is not set"
+  [[ -z "$HYPERVISOR" ]] && gitops_cluster_help "HYPERVISOR is not set"
```
whitfiea: why do we need the hypervisor param? Adding this would cause a breaking change, and it seems to be only needed to work out if we should created the efs cis driver or not? I would have thought the fact that the efs-csi-driver.yaml exists would be enough, similar to how other optional pieces work

whitfiea (changes requested):

image/cli/mascli/templates/gitops/appset-configs/cluster/instance/configs/ibm-mas-jdbc-config.yaml.j2
amitpandey0217: this is beacuse for rds we need it default we are not storing that under secret manager. you can check logic under gitops_mas_config. We have separated incluster, rds and default logic.

image/cli/mascli/functions/gitops_efs_csi_driver
dpdifferent1: Added defaults inside gitops-cluster.yml.j2

image/cli/mascli/functions/gitops_cluster
dpdifferent1: Hypervisor param was introduced in account params to restrict installation for clusters other than ROSA. However removed validation for hypervisor param from gitops_cluster function. Also, for gitops_efs_csi_driver function, we're checking if the `efs-csi-driver-role-arn` exists for the function to execute.

image/cli/mascli/functions/gitops_efs_csi_driver
whitfiea: the defaults need to be in this function as not everything uses the tekton pipelines (namely the reason why this function was added in the first place)

image/cli/mascli/functions/gitops_mas_config
```diff
@@ -913,10 +916,10 @@ function gitops_mas_config() {
         sm_update_secret $JDBC_CONFIG_SECRET_ID "{ \"jdbc_connection_url\": \"${JDBC_CONNECTION_URL}\", \"jdbc_instance_name\": \"${JDBC_INSTANCE_NAME}\", \"ca_b64\": \"${JDBC_CERTIFICATE_CONTENT_B64}\" }" "${TAGS}"
         echo_reset_dim "JDBC_INSTANCE_NAME ........................ ${COLOR_MAGENTA}$JDBC_INSTANCE_NAME"
         echo_reset_dim "JDBC_CONNECTION_URL ....................... ${COLOR_MAGENTA}${JDBC_CONNECTION_URL}"
+        export SECRET_KEY_JDBC_CONNECTION_URL=${JDBC_CONFIG_SECRET_ID}#jdbc_connection_url
+        export SECRET_KEY_JDBC_CERTIFICATE_CONTENT=${JDBC_CONFIG_SECRET_ID}#ca_b64
+        export SECRET_KEY_JDBC_INSTANCE_NAME=${JDBC_CONFIG_SECRET_ID}#jdbc_instance_name
       fi
-      export SECRET_KEY_JDBC_CONNECTION_URL=${JDBC_CONFIG_SECRET_ID}#jdbc_connection_url
-      export SECRET_KEY_JDBC_CERTIFICATE_CONTENT=${JDBC_CONFIG_SECRET_ID}#ca_b64
-      export SECRET_KEY_JDBC_INSTANCE_NAME=${JDBC_CONFIG_SECRET_ID}#jdbc_instance_name
```
whitfiea, line 919: The SECRET_KEY_JDBC_CONNECTION_URL and SECRET_KEY_JDBC_CERTIFICATE_CONTENT are no longer set when incluster-db2 is set, and so this would fail when we sent a blank path in the template

image/cli/mascli/functions/gitops_cluster
```diff
@@ -28,6 +28,7 @@ IBM Container Registry Entitlement (Required):
 
 Target Cluster (Optional):
       --cluster-url ${COLOR_YELLOW}CLUSTER_URL${TEXT_RESET}       Set to target a remote Kubernetes cluster (defaults to 'https://kubernetes.default.svc')
+      --hypervisor ${COLOR_YELLOW}HYPERVISOR${TEXT_RESET}         Hypervisor type for the cluster (e.g., 'rosa', 'fyre')
```
whitfiea: i thought the comment was to remove HYPERVISOR?

whitfiea (changes requested):

image/cli/mascli/functions/gitops_cluster
dpdifferent1: Not exactly.. Just removed the validation on it..

image/cli/mascli/functions/gitops_mas_config
dpdifferent1: Added default in the function. 
Added SECRET_KEY_JDBC_CONNECTION_URL and SECRET_KEY_JDBC_CERTIFICATE_CONTENT for in cluster db.

image/cli/mascli/functions/gitops_cluster
whitfiea: can we not remove it completely? As we don't need the extra parameter

dpdifferent1: Not only are we conditioning the CLI task execution based upon it, but also the Argo app creation through Gitops in order to restrict it from installing it on clusters apart from ROSA. Hence that's what me & Rob had agreed upon, considering we might utilise it for future use as well if required. Also let me know if it's still breaking since I'm trying to gather some context here.

dpdifferent1: Have removed the Hypervisor param from the CLI function gitops_cluster since it will not be used in Gitops. Now it's only being used in `gitops-mas-cluster` pipeline for executing `gitops-efs-csi-driver` function.

dpdifferent1: Removed Hypervisor implementation from the repo.. added field `efs_install_action` in `gitops-mas-cluster` pipeline for conditional install.

whitfiea (approved):

PR #2068, opened 2026-02-13
some tweaks to restore and update docs & guides 

sanjayprab:
dev->dev PR

- prompt mongodb storageclass override during restore
- update override options name for manage app db to be consistent with mongo options.
- update docs and guides

screenshot of interactive cli
Screenshot 2026-02-13 at 13 18 41

screenshot of params in openshift pipeline
Screenshot 2026-02-13 at 13 19 01


python/src/mas/cli/restore/app.py
```diff
@@ -146,11 +155,9 @@ def restore(self, argv):
             if self.args.mas_domain_on_restore is None:
                 self.promptForMASConfiguration()
 
-            if self.getParam("include_sls") == "true":
-                self.promptForSLSConfiguration()
```
sanjayprab, line 150: removed the conditions, as I want to display the prompt always in interactive. because, in scenario where user doesn't want to restore SLS/DRO instance from backup and still wants to chnage configuration slscfg. 

its wrong to add include_sls or include_dro condition here, because these prompts are for Suite-level SLS Configuration or Suite-level DRO Configuration

whitfiea (approved):

PR #2071, opened 2026-02-16
[patch] Add unit test for aiservice dev-mode installation

Bhautik-Vala:
Added comprehensive dev mode testing for mas aiservice-install command, mirroring the existing test structure in test/install/test_dev_mode.py


python/test/utils/install_test_helper.py
durera, line 1: Do we need to duplicate so much code for this?  We should be updating the existing helper to work for both aiservice and mas install testing, rather than seperate implementations that are mostly the same thing.

durera (changes requested):

python/test/utils/install_test_helper.py
Bhautik-Vala: yes, there is duplication - let me optimize the helper class for both type of installation.

durera (approved):

PR #2082, opened 2026-02-23
[patch] enable db2u_kind on cli and FVT

terc1997:
## Description
This PR adds db2u_kind to the pipeline to switch between Db2uCluster and Db2uInstance

## Unit tests
- https://github.com/ibm-mas/cli/pull/2082/changes#diff-03cf9150c7a02ca79cb0f3c49e27df63168426a160200997053d40d73237b37d
- https://github.com/ibm-mas/cli/actions/runs/22364978422/job/64728787317

## FVT Tests

**Test path**: /tests/db2instance2

**Screenshots**

image

image

image


python/src/mas/cli/install/settings/db2Settings.py
```diff
@@ -269,6 +269,16 @@ def configDb2(self, silentMode=False) -> None:
                     self.promptForString(" + Metadata Volume", "db2_meta_storage_size", default=self.getParam("db2_meta_storage_size"))
                     self.promptForString(" + Transaction Logs Volume", "db2_logs_storage_size", default=self.getParam("db2_logs_storage_size"))
                     self.promptForString(" + Backup Volume", "db2_backup_storage_size", default=self.getParam("db2_backup_storage_size"))
+
+                if self.yesOrNo("Select Db2 Custom Resource(CR)"):
```
durera: We don't want the user choosing the kind ... the control we have added is for dev mode only.

- Now: it's always db2ucluster
- Future: db2u_kind is controlled by the catalog you are using

durera: In other words, we need the cli to be able to control db2u_kind explicitly for development, while maintaining db2ucluster only normal usage - until we're ready to switch over and make db2uInstance be what customers are using (which will come after we have handled upgrade/migration etc).

durera (changes requested):

python/src/mas/cli/install/settings/db2Settings.py
terc1997: Okay. Check this commit https://github.com/ibm-mas/cli/pull/2082/changes/6c9030790f32027693fdb539752a92c17105db5b. Pretty much if it's not `dev-mode` it's `db2ucluster`.

durera (approved):

PR #2099, opened 2026-03-04
[minor] Support db2 v12 upgrade

terc1997:
## 1. Description
To support Db2 v12 upgrade, it requires to change `mas update` command to allow the upgrade according with the user input. So, for that support it will be included:
- `--db2-v12-upgrade` and `--db2-license-file` flag in the `mas update` command line
- `db2_v12_upgrade` and `db2_license_file` flag in `mas_update` Tekton task, backed up by a change in the ansible-devops to carry `db2_v12_upgrade`
- Update unit tests 

Also in this PR it will be delivered the support for Db2 License file. To support that:
- Create `db2LicenseFile()` which has the same features as `slsLicenseFile()`.
- It was created a new workspace `shared-db2` that mounts `pipeline-db2-license` secret. 

## 2. Dependencies
- https://github.com/ibm-mas/ansible-devops/pull/2140 
- https://github.com/ibm-mas/python-devops/pull/204 
### 3.1. CLI Help
image

#### Without `--db2-v12-upgrade`
image

#### Without `--db2-license-file`
image

#### Normal flow
image

### 3.3. Interactive
image

### CLI Help
image

### Interactive
image
image

### Non Interactive
image
image

### Pipeline with License
image
image
image
image

### Pipeline without License
image
image
image

### FVT Tests
 - db2upgrade3
 - db2instance
<hr/>


durera (commented):
In the screenshots, it doesn't show how the user will be prompted for the db2 activation key in interactive or non-interactive ... I assume that in there though, otherwise upgrade would take them to a trial license?  And we need to prevent the update being started in the event that they don't provide the activation key -- even if they think they want to move to the trial license, I just don't want that headache when it means a prod system goes offline in 30 days :D 

We can let people install NEW without a license, but for the update/db2 v12 upgrade I think we need to ensure that there's no "downgrade" from full to trial as part of this action

terc1997:
> In the screenshots, it doesn't show how the user will be prompted for the db2 activation key in interactive or non-interactive ... I assume that in there though, otherwise upgrade would take them to a trial license? And we need to prevent the update being started in the event that they don't provide the activation key -- even if they think they want to move to the trial license, I just don't want that headache when it means a prod system goes offline in 30 days :D
> 
> We can let people install NEW without a license, but for the update/db2 v12 upgrade I think we need to ensure that there's no "downgrade" from full to trial as part of this action

Okay. I didn't enforce the license for db2 v12 upgrade. Let me update CLI and ansible-devops to enforce the presence during the upgrade.

terc1997:
Added license enforcing when updating from v11 to v12.

durera (dismissed):

PR #2108, opened 2026-03-12
[patch] FVT - sync window not deleted properly

mnivedithaa:
Issue: [MASCORE-12682](https://jsw.ibm.com/browse/MASCORE-12682)

## Description
Sync window for fvtsaas2 is not deleted properly
This is mainly because of the difference in argocd cli version(v3.2.0+66b2f30) and server version(v3.0.23+7cedae7)

### DIFF
--- a/tekton/src/tasks/gitops/gitops-mas-fvt-preparer.yml.j2
+++ b/tekton/src/tasks/gitops/gitops-mas-fvt-preparer.yml.j2
@@ -736,7 +736,7 @@ spec:
 
         # Remove argocd window
         argocd_login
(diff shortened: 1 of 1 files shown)


whitfiea (changes requested):
would this break if we updated the argocd cli ?

mnivedithaa:
It depends on what changes are included in the latest argocd cli
In Argo CD 3.1+ a change was made to the sync windows output structure.
The newer CLI prints an extra field (like ManualSync / Enabled state), but the 3.0 server API still returns the older structure.
So the CLI formatter receives an extra argument and Go prints %!(EXTRA string=Disabled)

whitfiea (approved):
Assuming that if the EXTRA isn't there then this should still work so approving it

PR #2116, opened 2026-03-19
[minor] Support March Interim Catalog Update: 260318

anilprajapatiibm:
Catalog: https://github.ibm.com/maximoappsuite/ibm-maximo-operator-catalog/pull/934

Address issue with kubernetes client and no_proxy
Emergency fix for operators explicitly installing mas-utils as this dependency was also including the kubernetes v35 library. This fix in the Dockerfile should be removed on the next 9.1.x release of ibm-mas.
Downgrade the Python kubernetes client to 33.1.0 as a temporary workaround for a proxy-related regression introduced in newer releases. Upstream issues report that versions 34.1.0+ can mishandle no_proxy, causing Kubernetes API requests that should bypass a configured proxy to be routed through it instead. In our OpenShift Ansible operator environment, this affects in-cluster API access when cluster-wide proxy variables are present. This change restores the previously working client version while we investigate if a longer-term fix: explicitly configuring kubernetes.core with K8S_AUTH_PROXY and K8S_AUTH_NO_PROXY is required rather than relying on inherited standard proxy environment variables.


python/src/mas/cli/install/catalogs.py
whitfiea, line 1: why are we removing the 260313 catalog rather than just adding a new one? The problem reported isn't seen by everyone, so it is still a valid catalog

whitfiea (changes requested):

python/src/mas/cli/install/catalogs.py
anilprajapatiibm: We have concluded on past, that we will replace interim releases with new interim or official release
and only one catalog fix will be given per month to reduce the confusion,

and for 260313 catalog they can still use the older release of CLI

whitfiea (approved):

PR #2133, opened 2026-03-27
[minor] Add ODH to RHOAI migration support for mas update

Divyesh-Khokhar:
### Description
Implements automatic detection and migration from OpenDataHub (ODH) to Red Hat OpenShift AI (RHOAI) during MAS update operations.

### Jira
- https://jsw.ibm.com/browse/MASAIB-1929
- https://jsw.ibm.com/browse/MASAIB-2043

### Testing
New installation:
image
image

MAS Update:
Screenshot 2026-04-02 at 7 21 43 PM
Screenshot 2026-04-08 at 7 45 37 PM
Screenshot 2026-04-08 at 7 46 23 PM


karol-czarnecki (dismissed):
lgtm

tekton/src/tasks/dependencies/ibm-catalogs.yml.j2
```diff
@@ -69,6 +79,18 @@ spec:
       image: quay.io/ibmmas/cli:latest
       imagePullPolicy: $(params.image_pull_policy)
       workingDir: /workspace/configs
+    
+    - name: aiservice-rhoai-install
+      when:
+        - input: "$(params.odh_to_rhoai_migration)"
+          operator: in
+          values: ["true"]
+      command:
```
whitfiea: we would want this role to run if deploying AIService rather than odh_to_rhoai_migration being true? I don't think we need odh_to_rhoai_migration at all. As the ibm_catalogs task doesn't seem the right place for this?

The uninstall of odh should be done if odh is detected.

whitfiea (dismissed):

tekton/src/tasks/dependencies/ibm-catalogs.yml.j2
Divyesh-Khokhar: @whitfiea We have added [detection logic](https://github.com/ibm-mas/ansible-devops/pull/2153/changes#diff-697cbd23249d35f2f89a61f67542e7fe3606dbe2536ef433138b31319971d399R1-R23) for ODH in the `aiservice_rhoai` role, and if it is detected, we [uninstall](https://github.com/ibm-mas/ansible-devops/pull/2153/changes#diff-44017a28a6adc9c9fca3d28027bff4f93243f1d330a40495d0bcea0b2bfe6586R48) ODH from there.

So, to display it during the CLI `mas-update` and to identify whether we should run the `aiservice_rhoai` role or not, I had kept the `odh_to_rhoai_migration` variable.

Should I remove the `odh_to_rhoai_migration` variable and simply run the `aiservice_rhoai` role if AI service is deployed as you mentioned? And just keep the `detectODH()` function in the CLI to display a notice about ODH to RHOAI migration?

whitfiea: For the update we can keep odh_to_rhoai_migration as the condition. i hadn't looked at the ansible code to see that it would uninstall odh. The question i still there about why this is in the catalog task ad not in its own task in the update pipeline

Divyesh-Khokhar: Sure @whitfiea , I’ll keep the `odh_to_rhoai_migration` condition. and Should we create a separate task, i.e., `rhoai.yml.j2`, under `tasks/dependencies` and move this logic there?

Divyesh-Khokhar: @whitfiea I’ve made the required changes, tested them, and attached the test results. Since we already have a task named rhoai under aiservice that’s used in mas-install, I’ve named the new task `rhoai-migration.yml.j2`. Could you please check and let me know if that looks fine?

sanjayprab (approved):

PR #2152, opened 2026-04-02
[minor] Added redis changes for collaborate add-on

nehasangwai-ibm:
https://github.com/ibm-mas/ansible-devops/pull/2181


python/src/mas/cli/install/settings/manageSettings.py
```diff
@@ -176,6 +176,10 @@ def manageSettingsComponents(self) -> None:
                     self.params["mas_appws_components"] += ",aip=latest"
                 if self.yesOrNo(" - Vegetation Management"):
                     self.params["mas_appws_components"] += ",vegm=latest"
+                if self.yesOrNo(" - Collaborate"):
```
whitfiea, line 179: This is 9.2.x only or 9.2.x-feature? Can we only prompt this if the channel supports it?

whitfiea (changes requested):

python/src/mas/cli/install/settings/manageSettings.py
nehasangwai-ibm: This is 9.2.x & onwards, similar to vegM, AIP that we have

image/cli/install/ibm-mas_devops.tar.gz
whitfiea, line 1: this file needs to be removed

image/cli/install/mas_devops.tar.gz
whitfiea, line 1: this file needs to be removed

.github/workflows/verify-cli-pr.yml
```diff
@@ -16,16 +16,3 @@ jobs:
...
-          RC=0
-          if [[ -e $GITHUB_WORKSPACE/image/cli/install/ibm-mas_devops.tar.gz ]]; then
-            echo "Found local Ansible collection($GITHUB_WORKSPACE/image/cli/install/ibm-mas_devops.tar.gz) in this PR, Local collection is not allowed in the master branch."
-            RC=1
-          fi
-          if [[ -e $GITHUB_WORKSPACE/image/cli/install/mas_devops.tar.gz ]]; then
-            echo "Found local Python package ($GITHUB_WORKSPACE/image/cli/install/mas_devops.tar.gz) in this PR, Local package is not allowed in the master branch."
-            RC=1
-          fi
-          exit $RC
```
whitfiea, line 31: this needs to be re-added in.

python/src/mas/cli/install/settings/manageSettings.py
whitfiea: So what happens if I am installing 9.1 and I say yes to this prompt? We need to tell the user that this is only for certain versions. The aip and vegm was added incorrectly without that check/help

whitfiea (dismissed):

python/src/mas/cli/install/settings/manageSettings.py
nehasangwai-ibm: Yes Andrew, this collaborate add-on is only from 9.2.x & onwards, this will be conveyed in note.

whitfiea: sorry, what note do you mean?

nehasangwai-ibm: @whitfiea  Currently, there are no customers using earlier versions of the Collaborate application. However, we will publish a technical note to formally communicate that starting from version 9.2.x onwards, Collaborate will be treated as an add-on component within Manage.

whitfiea: I think you are misunderstanding what I am saying. As a user if I try to install Manage 9.1.x then I am still prompted to install colloborate or not. At the very least we should have the prompt state that this is applicable to only 9.2.x, and a better solution would be to not prompt at all if using 9.1x or 9.0.x etc

nehasangwai-ibm: Got it, @whitfiea. We can look into updating the prompt so that it clearly applies only to version 9.2.x onwards.
For now, can we proceed with merging this for the April FC?
Also, just to clarify—currently, Collaborate is not installed by default. It is only installed if the user explicitly selects ‘y’ when prompted for Manage add-on components.

please find the snippet : 

Screenshot 2026-03-11 at 2 44 58 PM

sanjayprab:
I raised the same concern that Andrew raised. the collaborate option in managesettings.py is supposed to be for 9.2 only and also after April. I had a call with @salma-source and @nehasangwai-ibm ,I requested that we should have a MAS version & catalog check condition surrounding this new option. But due to some time constraints and dcut, they  are unable to make the changes now and have got approval from Yin to have a technote/known issue for Users to not select this new 'collaborate' option unless they install April FC. So I am proceeding with approving this PR considering the approval they have got. @salma-source will create and track bug issue to fix this asap.

sanjayprab (approved):

PR #2154, opened 2026-04-03
[minor] add cli support for mas without cluster role [MASCORE-13444]

dixitgsathwara:
## Description
This PR introduces a new mas pre-install CLI command for pre-install RBAC setup and adds permission mode support to `mas install` and `mas aiservice-install`, allowing users to choose between three RBAC configurations during installation: `cluster`, `namespaced`, and `minimal`

- Add a new mas pre-install command for  RBAC setup
- permission mode support in mas install and mas aiservice-install
- Integrated pre-install RBAC files into CLI Docker image build process

## Guidance for Reviewers
- Before merging this PR, the following PRs must be merged first:
   https://github.com/ibm-mas/ansible-devops/pull/2185
   https://github.com/ibm-mas/python-devops/pull/261

## Jira Issue
- [MASCORE-13444](https://jsw.ibm.com/browse/MASCORE-13444)
- [MASCORE-12621](https://jsw.ibm.com/browse/MASCORE-12621)

## Testing 

- mas `pre-install`
  image

- when user do not have cluster admin permissions to apply pre-install rbac 
   
  image
  image

- Inline cli pre-install rbac apply before launching install pipeline
   image

 
  
image


python/src/mas/cli/install/app.py
```diff
@@ -85,6 +85,32 @@ def wrapper(self, *args, **kwargs):
 
 
 class InstallApp(BaseApp, InstallSettingsMixin, InstallSummarizerMixin, ConfigGeneratorMixin, installArgBuilderMixin):
+
+    def setSelectedAppsParam(self):
+        selectedApps = ["core"]
```
IanBoden, line 91: I'm not sure we would always want core to be selected, if you install aiservice you don't need core

durera: What are we ultimately doing with this new `mas_selected_apps` param?  It seems to be a duplication (via combination) of existing parameters.  Don't we already have all this information available to us jsut as easily by passing the various `_version` params into the new task.  Not a fan of creating new params that duplicate information in other params, it creates a tekton API that is difficult to use, and just because we put a wrapper around that API with the CLI doesn't mean we should do that.

If the suite-rbac task needs a comma separated list of apps, IMO it should have logic that builds it based on the existing params ...

e.g. 
```
if mas_instance_id != "":
  # add core
  if mas_app_version_iot != "":
    # add iot
  # etc
if aiservice_instance_id != "":
  # etc
```

tekton/src/pipelines/mas-install.yml.j2
```diff
@@ -238,19 +238,30 @@ spec:
           operator: notin
           values: [""]
 
-    # 5.4 Suite Installation
+    # 5.4 Apply RBAC based on permission mode
```
durera: My suggestion would be to do this as the very first thing we do ... even before the catalog.  IMO it's the logical order:

- validate health of cluster (pre-verify)
- install rbac
- start installing

At the moment we set up the rbac mid-install.  Even though we are installing it before the suite stuff starts so technically this isn't an issue, I think in terms of separating the activities here and having a pipeline that reflects how you would do this as a human, we should have this at the start of the pipeline

tekton/src/pipelines/taskdefs/core/suite-rbac.yml.j2
durera, line 1: I think we should position this more as `rbac` (without the suite part) ... `suite-x` stages/tasks implies we are doing things just in scope of MAS, but e.g. already we would be doing non-mas rbac too (aiservice), and in future - if not already - we may have to/want to pre-install non-optional RBAC for other products in this same manner ... e.g. atm a bunch of RBAC stuff is done in the mongo role.

Look at this as "we are setting up RBAC for everything we are about to install" rather than "we are setting up RBAC for mas (core + iot + manager + etc)" .. because I think that will be where we end up.

durera (changes requested):

durera (approved):
As discussed in Slack

PR #2159, opened 2026-04-08
[patch] New Mobile FVT Tasks: fvt-component and Collaborate desktop task

leo-miran:
New tasks:
- `fvt-component` for the pipeline : so that tasks can be skipped directly from pipeline
- `Collaborate`: New desktop fvt test for Collaborate Application

Test Results:
- Mobile pytest when conditions test
image

- Mobile testng new task and when conditions test
image


IanBoden (approved):
looks good

PR #2162, opened 2026-04-09
[patch] mirror tenant operator

boonware:
## Issue
https://jsw.ibm.com/browse/MASAIB-1997

## Description
Add the new AI Service tenant operator to image mirroring. This operator is mandatory for AI Service 9.2+. The operator subscription and CASE are named `ibm-aiservice-tenant`


python/src/mas/cli/mirror/app.py
```diff
@@ -745,3 +748,10 @@ def mirror(self, argv):
                 )
 
             print_formatted_text(HTML("\n<B>✅ Mirror operation completed</B>"))
+
+    def _is_unsupported_package(self, version: str, packageName: str) -> bool:
+        unsupported = False
+        if packageName == "ibm-aiservice-tenant" and version.startswith("9.1."):
+            logger.info("Package 'ibm-aiservice-tenant' only supported for 9.2.x or higher")
```
IanBoden: so from the code it seems as though if someone trys to request ibm-aiservice-tenant on 9.1 we will just skip it and print out this info, seems like it should at least be a warning, I think it should also say explicitly that it wont be mirrored. So something like:
logger.warning("Skipping mirroring package 'ibm-aiservice-tenant' due to unsupported version, only supported for 9.2.x or higher")

IanBoden (approved):
looks good

PR #2232, opened 2026-04-30
[patch] fix ai svc slack notification

anilprajapatiibm:
ai service install pipeline using aiservice-{instance-id}-pipelines,
Slack notification was trying to use mas-{instance-id}-pipelines, due to that it was failing to fetch the namespace

passing namespace into the notification call, to directly handle the namespace

https://jsw.ibm.com/browse/MASAIB-2186

Python-devops: https://github.com/ibm-mas/python-devops/pull/313

Test:
image
image


terc1997 (commented):
LGTM. Address the review at https://github.com/ibm-mas/python-devops/pull/313 and update the methods.

terc1997 (approved):
LGTM

PR #2237, opened 2026-05-01
[PATCH] add service mesh stuff for the apps

IanBoden:
Move the mas-use-service-mesh parameter from suite-install into cli-params so it can be used by the app installs as well. 

Add it to all the templates and fix the mas-install pipeline to correctly use the iot and monitor template in 9.2 rather than having it copied directly into the template, unlike the other templates they do not contain the "name" or "when" as those need to change based on the mas version.

Tested via personal fvt in conjuction with https://github.com/ibm-mas/ansible-devops/pull/2229


tekton/src/pipelines/taskdefs/apps/iot-app.yml.j2
durera, line 1: Are these removals intentional?

tekton/src/pipelines/taskdefs/apps/monitor-app.yml.j2
durera, line 1: Are these removals intentional?

tekton/src/pipelines/mas-install.yml.j2
```diff
@@ -334,39 +335,7 @@ spec:
     # Note: In the case of manage using the shared Db2 instance only start the
     # IoT install *after* any Manage database preparation has completed
     - name: app-install-iot-after-monitor
-      timeout: "0"
```
durera, line 337: Are these removals intentional?

tekton/src/pipelines/mas-install.yml.j2
```diff
@@ -396,31 +366,7 @@ spec:
     # 10. Install & Configure Monitor (Monitor < 9.2.0 - legacy, installs after IoT)
     # -------------------------------------------------------------------------
     - name: app-install-monitor-after-iot
-      timeout: "0"
```
durera, line 399: Are these removals intentional?

tekton/src/pipelines/taskdefs/apps/monitor-app.yml.j2
IanBoden: yes the "name" and "when" are now in the mas-install.yaml as they change based on order of iot and monitor.

tekton/src/pipelines/mas-install.yml.j2
IanBoden: yes everything other than "name", "when", and "runAfter" should be in the template. 

The timeout was being set in the case of installing monitor before iot but not when installing monitor after iot. Couldn't see the timeout being useful and was only specified when doing iot after monitor so chose to remove it to match the established 9.1 settings

tekton/src/pipelines/mas-install.yml.j2
IanBoden: yip see above

tekton/src/pipelines/taskdefs/apps/iot-app.yml.j2
IanBoden: yes the "name" and "when" are now in the mas-install.yaml as they change based on order of iot and monitor.

durera (approved):

PR #2243, opened 2026-05-05
[minor] Enable ObjectStorageCfg provisioning for MVI application (COS mapping + IAM policy update)

nikitakumble:
### issue:
https://jsw.ibm.com/browse/MASCORE-13155

### Description:
This PR enables provisioning of `ObjectStorageCfg` for the MVI application. 
`ibm-iac-main.tf.j2` is updated to include the `s3:ListAllMyBuckets` IAM policy, which is required for bucket listing in `ObjectStorageCfg`.

`utils` function is also updated to correctly handle quoted values, ensuring proper parsing when secret values contain double quotes.

### Related PR:
https://github.ibm.com/maximoappsuite/saas-envs-starter/pull/159
https://github.ibm.com/maximoappsuite/saas-tekton/pull/226


whitfiea (changes requested):
I am not sure I understand why this is needed. The terraform already creates the right version as long as you set the format correctly https://github.ibm.com/maximoappsuite/mas-iac-aws-s3/blob/master/module/main.tf#L296-L297. The version that has the accesskey in the secret name is when it is a `manage` version of s3 secret. The `cos` version aka objectstoragecfg should already set it correctly

nikitakumble:
> I am not sure I understand why this is needed. The terraform already creates the right version as long as you set the format correctly https://github.ibm.com/maximoappsuite/mas-iac-aws-s3/blob/master/module/main.tf#L296-L297. The version that has the accesskey in the secret name is when it is a `manage` version of s3 secret. The `cos` version aka objectstoragecfg should already set it correctly

Got it now. The issue was in the IAC Terraform template (line 419) where cos_secret_format = false, which prevented the module from creating the secret with the correct ObjectStorageCfg format (username/password/info).
Fixed by setting cos_secret_format = true for the MVI bucket. The transformation logic in gitops_mas_config has been reverted since the Terraform now creates the secret in the proper format from the start.

image/cli/mascli/functions/internal/utils
```diff
@@ -426,10 +426,27 @@ def resolve_path(obj, path):
...
+    else:
+        return obj
+
 data = unwrap_singleton_lists(load_hcl(file_path))
 
 try:
     result = resolve_path(data, dot_path)
+    # Strip quotes from result (handles strings, lists, dicts recursively)
+    result = strip_quotes_recursive(result)
+
```
whitfiea, line 449: do we still need these changes?

image/cli/mascli/templates/gitops/appset-configs/cluster/instance/ibm-iac-main.tf.j2
```diff
@@ -416,7 +416,7 @@ locals {
           access_level         = "rw"
           s3_bucket_arn        = "arn:aws:s3:::${var.s3bucket_name_prefix_mvi}-mvi"
           secret_suffix        = "cos"
-          cos_secret_format    = false
+          cos_secret_format    = true
           manage_secret_format = true
```
whitfiea: the manage_secret_format needs to be false

whitfiea (changes requested):

image/cli/mascli/functions/internal/utils
nikitakumble: Yes, this function is needed. It helps extract correct values from Terraform cluster files. Without `strip_quotes_recursive`, values were coming with extra double quotes (e.g., ""vpc-12345"" instead of "vpc-12345"), causing failures, particularly in the `gitops_iac` function which calls `get_tf_value` multiple times for VPC, CIDR, and other values.

image/cli/mascli/templates/gitops/appset-configs/cluster/instance/ibm-iac-main.tf.j2
nikitakumble: Updated.

sumitjain-1901 (dismissed):

whitfiea (approved):

PR #2247, opened 2026-05-06
[minor] Add changes for installing AI Assistant and AI Service on same cluster

joshiricha444:
Jira: [](https://jsw.ibm.com/browse/MASR-8107)

This PR enables installation of AIAssistant through the CLI and covers the following scenarios:

**Scenario 1:**
AI Service is installed on a different cluster. In this case, AI Service details must be provided explicitly to configure Meta.

**Scenario 2:**
AI Service, Core, Manage, and AIAssistant are installed on the same cluster. In this setup, the `aicfg` automatically discovers AI Service details and uses them to create the AIAssistant CR.

image (1)
image
scenario 1- where aiservice is not installed on same cluster
image
scenario 2- where aiservice is installed on same cluster(so AIAssistant will pick the aiservice details)
image
image


tekton/src/tasks/gencfg-aiservice.yml.j2
```diff
@@ -0,0 +1,146 @@
...
+              sslEnabled: true
+              credentials:
+                secretName: "${MAS_INSTANCE_ID}-usersupplied-ai-creds-system"
+            assistantAgent:
+              enabled: true
+          certificates:
+            - alias: aiservice
+              crt: |-
+        $(echo "${CERTIFICATE}" | sed 's/^/        /')
+        EOF
```
whitfiea, line 140: Although the cli generates the template in the `gencfg.py` using the template in aicfg.yml.j2 we seem to ignore that and just set this version, unless i have misunderstood?

python/src/mas/cli/gencfg.py
```diff
@@ -60,6 +60,52 @@ def generateJDBCCfg(self, instanceId: str, scope: str, destination: str, appId:
...
+            mas_instance_id=instanceId,
+            mas_workspace_id=workspaceId,
+
+            cfg_display_name=name,
+
+            ai_url=url,
+            ai_tenant_id=tenantId,
+            ai_apikey=apikey,
+            ai_enabled=enabled,
+            meta_agent_enabled=metaAgentEnabled,
```
whitfiea: can we change it to aiassistant?

python/src/mas/cli/install/argParser.py
```diff
@@ -114,6 +114,12 @@ def isValidFile(parser: argparse.ArgumentParser, arg: str) -> str:
     required=False,
     help="AI Service Instance ID"
 )
+masArgGroup.add_argument(
+    "--configure-AIAssistant",
```
whitfiea: we shouldn't have the uppercase AIA here, either `--configure-aiassistant` or `--configure-ai-assistant`

whitfiea (changes requested):

tekton/src/tasks/gencfg-aiservice.yml.j2
aniketchoudhary03: Thanks for the comment.

We have two scenarios here:

When installing with AI Service on the same cluster, the Meta CR is generated automatically after AI Service is installed. In this case, the template from gencfg.py (using aicfg.yml.j2) is effectively used, and the required details are picked up automatically.
When AI Service is on a different cluster, we explicitly set the version and required details to configure Meta correctly, since they are not auto-discovered.

So the explicit version setting is mainly to handle the external AI Service scenario, while the template-driven approach applies to the in-cluster setup.

python/src/mas/cli/install/settings/aiSettings.py
```diff
@@ -0,0 +1,280 @@
...
+                'tenant_id': aiservice_tenant_name,
+                'apikey': apikey,
+                'certificate': certificate
+            }
+
+        except Exception as e:
+            print_formatted_text(f"\n✗ Error during auto-detection: {str(e)}\n")
+            return {}
+
+    def _generateAiCfgWithValues(self, instanceId: str, scope: str, destination: str, workspaceId: str,
```
whitfiea: i don't think _generateAiCfgWithValues is used anywhere

python/src/mas/cli/install/settings/aiSettings.py
```diff
@@ -0,0 +1,280 @@
...
+            ...
+
+        # Methods from ConfigGeneratorMixin or InstallSettingsMixin
+        def selectLocalConfigDir(self) -> None:
+            ...
+
+        def generateAiCfg(self, instanceId: str, scope: str, destination: str, workspaceId: str = "") -> None:
+            ...
+
+    def _detectAiServiceFromCluster(self) -> Dict[str, str]:
```
whitfiea: I don't think _detectAiServiceFromCluster is used anywhere

whitfiea (changes requested):

tekton/src/tasks/gencfg-aiservice.yml.j2
whitfiea: ok, I see the difference now, there is an if else block based on if aiservice is to be installed. If it is then configure_aiassistant gets set to pipeline, else it renders the config in the config dir to be applied for an external aiservice.

aniketchoudhary03: yes, Thanks

python/src/mas/cli/install/settings/aiSettings.py
aniketchoudhary03: removed

python/src/mas/cli/install/settings/aiSettings.py
aniketchoudhary03: removed

python/src/mas/cli/install/argParser.py
aniketchoudhary03: change it "--configure-ai-assistant"

python/src/mas/cli/gencfg.py
aniketchoudhary03: changed to ai_assistant_enabled=aiAssistantEnabled and  aiAssistantEnabled = self.yesOrNo("Enable AI Assistant Agent (AI assistant for MAS)")

whitfiea (dismissed):

whitfiea (approved):

PR #2248, opened 2026-05-06
[patch] MAS and Manage RBAC Permissions in Install pipeline

unnati-solanki-git:
https://jsw.ibm.com/browse/MASCORE-13654

image

image


image/cli/mascli/functions/must_gather
```diff
@@ -224,14 +224,6 @@ function mustgather() {
   echo "For help reviewing the content of the must gather refer to:"
   echo "https://www.ibm.com/support/pages/how-review-maximo-application-suite-must-gather"
 
-  if [ "$(oc whoami 2>/dev/null)" == "" ] ; then
```
durera, line 227: Even if we don't require use to be a cluster admin (second check), we still want the first check (that they are logged into the cluster) ?

rbac/install/pipeline/clusterrole.yaml
```diff
@@ -11,6 +11,9 @@ subjects:
   - kind: ServiceAccount
     name: mas-{{ mas_instance_id }}-install-pipeline
     namespace: mas-{{ mas_instance_id }}-pipelines
+  - kind: ServiceAccount
```
durera: why do we need to bind to this service account as well?

If user is using the OOTB "pipeline" account it means they have chosen to go for the "just set me up the default  account with cluster admin permissions still doesn't it?

image/cli/mascli/functions/must_gather
unnati-solanki-git: Done

rbac/install/pipeline/clusterrole.yaml
unnati-solanki-git: Thanks for the noticing! This was added to test must-gather related changes, but forgot to clean up. Removed now!

durera (approved):

PR #2257, opened 2026-05-08
[patch] Opendatahub update mirror images

karol-czarnecki:


python/src/mas/cli/mirror/config.py
```diff
@@ -14,6 +14,7 @@
     ("Optional Dependencies", "amlen", "amlen", "amlen_extras_version"),
     ("Optional Dependencies", "aiservice", "ibm-aiservice", "aiservice_version"),
     ("Optional Dependencies", "aiservice", "ibm-aiservice-tenant", "aiservice_version"),
+    ("Optional Dependencies", "odh", "opendatahub", "odh_version"),
```
durera: Add this into the `aiservice` group so that it's mirrored as part of choosing to mirror aiservice packages

durera (changes requested):

python/src/mas/cli/mirror/config.py
karol-czarnecki: fixed

durera:
Not needed - this has been merged to master in another PR

PR #2266, opened 2026-05-12
[patch] Adding MAS_CHANNEL env var to suite-dns 

vpatel9305:
we removed some functionality in IoT from 9.2 and hence we are removing the routes too. These are applicable till 9.1
Hence DNS Entries requires version control so it wont be there from 9.2 MAS

9.1.x
image

9.2.x
image


authorized_entitlement.lic
```diff
@@ -0,0 +1,183 @@
+SERVER sls-rlks-0.rlks 0242ac110002 27000
```
dixitgsathwara: why we need this ?

entitlement.lic
```diff
@@ -0,0 +1,183 @@
+SERVER sls-rlks-0.rlks 0242ac110002 27000
```
dixitgsathwara: why we need this file

dixitgsathwara (dismissed):

authorized_entitlement.lic
vpatel9305: We do not need it. Removed

entitlement.lic
vpatel9305: Removed this too

IanBoden (approved):

PR #2268, opened 2026-05-12
[patch] Added the facilities in mas-upgrade pipeline

sdoshi96:
## 1. Description
While monitoring the custom Day 2 pipeline FVT run, we observed that the facilities upgrade task is missing from the `mas-upgrade` pipeline flow. As a result, the facilities upgrade step was not executed in any of the scheduled or custom FVT Day 2 runs. Hence, added the facilities in `mas-upgrade` pipeline

## 2. Issue
- https://jsw.ibm.com/browse/MASREF-5096
 
## 3. Tests

image

image


terc1997 (approved):
LGTM

PR #2279, opened 2026-05-14
[minor] Add param for AI Service deprovisioning in gitops

boonware:
## Issue
MASAIB-2266

## Description
New parameter to enable tenant deprovisioning for AI Service in the GitOps flow.

## Test Results
Confirmed via testing in the Tekton pipeline that the change is rendered correctly and the correct output params file for ArgoCD is generated.


tekton/src/tasks/gitops/gitops-aiservice-tenant.yml.j2
```diff
@@ -23,6 +23,9 @@ spec:
       type: string
     - name: github_host
       type: string
+    - name: aiservice_enable_deprovisioning
+      type: string
+      default: "false
```
jasmin-30: Quotes left unclosed.

jasmin-30 (changes requested):

tekton/src/tasks/gitops/gitops-aiservice-tenant.yml.j2
boonware: Thanks, good catch. I had fixed this in Tekton, forgot to do it here.

jasmin-30 (approved):

whitfiea (approved):

PR #2295, opened 2026-05-20
[patch] fix errors and reword servicemesh prompts

stonepd:
### Description
While running mas install for issue https://jsw.ibm.com/browse/MASCORE-13906 
I got a couple of annoying stack trace dumping errors, 
1 - when running `mas install ` in non interactive mode and I have not logged on to the kubernetes cluster
2 - when entering a blank "Custom channel for manage"

This fixes those by
1 - checking for a valid client in lookupTargetArchitecture, also change subsequent warning from ` The Kubernetes dynamic Client is not available` to a more friendly `Not successfully connected to a Kubernetes cluster`
2 - validating that any entered value for application Custom channels are not empty

The service mesh MAS text has been modified and the manual route management mode is hidden in the install cli.

### Testing
Tested by running the built CLI


python/src/mas/cli/install/app.py
```diff
@@ -1016,23 +1016,19 @@ def _checkIngressControllerPermissions(self, controllerName="default"):
...
-        if self.showAdvancedOptions:
-            self.printH1("Configure Routes Manually")
-            self.printDescription(["Disable automatic route creation."])
-            self.yesOrNo("Disable Route Creation", "mas_manual_route_mgmt")
-        else:
-            self.setParam("mas_manual_route_mgmt", "false")
-
     @logMethodCall
     def configServiceMesh(self) -> None:
         if self.showAdvancedOptions:
```
IanBoden: we need a "and 9.2.0 or later" bit added here

IanBoden (changes requested):
just 1 small change needed

python/src/mas/cli/install/app.py
stonepd: done

IanBoden (approved):

PR #2300, opened 2026-05-21
[patch] Fix core tests pipeline failure

unnati-solanki-git:
https://jsw.ibm.com/browse/MASCORE-14243

Tested locally -

[local-test-logs.txt](https://github.com/user-attachments/files/28110236/local-test-logs.txt)

image


terc1997 (commented):
LGTM. Quick question, how does the dashboard behave in case of `onError: continue` ? Does it log everything ? If so, I think we should add the `onError: continue` in all FVT tests to not block the whole suite test.

unnati-solanki-git:
> LGTM. Quick question, how does the dashboard behave in case of `onError: continue` ? Does it log everything ? If so, I think we should add the `onError: continue` in all FVT tests to not block the whole suite test.

I did a small check with similar config to understand exact behaviour. The task will still report failure but will not fail the pipeline and continue with next task in sequence. 

More details here - https://ibm-mas.slack.com/archives/C0A6RLMP614/p1779390573568699?thread_ts=1779361593.751849&cid=C0A6RLMP614

PR #2303, opened 2026-05-22
[patch] Add pre-checks and apply pre-install RBAC for upgrade for 9.2

Aneri11U:
## Description
- In case of 9.2 + , check the permissionMode and if minimal found then block the upgrade otherwise apply the pre-install RBAC for the detected permissionMode. 
- For, 91 to 92 upgrade - this takes permissionMode as cluster by default
- Fix the syntax error for sync-upgrade task

## Jira Issue
- https://jsw.ibm.com/browse/MASCORE-13445

## Testing
- Tested by upgrading from 9.1.x to 9.2.x-dev for core and manage
image
On cluster the roles gets created
image
Upgrade for core and manage got successful
image
sync-upgrade failed due to passing ${CONFIGMAP_VALUE} as `Completed #-------  ` was not under quotes so giving bad flag error while creating configmap:
image

After the fix:
image


python/src/mas/cli/cli.py
```diff
@@ -226,7 +226,6 @@ def __init__(self) -> None:
         }
 
         self.upgrade_path: Dict[str, str] = {
-            "9.1.x": "9.2.x-feature",
```
durera, line 229: is this removal intentional?

Aneri11U: No I think while testing I added the path to 9.1.x to 9.2.x-dev, might have removed this existing one accidentally. Thanks for pointing this out! 
I have put this back.

durera (approved):

PR #2312, opened 2026-05-26
[patch] corrected the pods-only parameter in doc example

dushyantbangal:
The parameter is `--pods-only`, as mentioned in the [doc](https://github.com/ibm-mas/cli/blob/f02864d64f6a6463852a709b9797fa21b661e9df/docs/commands/must-gather.md#L16) and in the [code](https://github.com/ibm-mas/cli/blob/f02864d64f6a6463852a709b9797fa21b661e9df/image/cli/mascli/functions/must_gather#L159)

However, the example in the doc called it `--pods_only`. I have corrected it.


whitfiea:
code no longer exists

PR #2317, opened 2026-05-27
[patch] add default routing mode path for 9.2+ installation

dixitgsathwara:
## Description

- currently cli doesn't contain the default routing mode for mas 9.2+ installation so due to that it's not set the mas_routing_mode variable but `ansible-devops` have condition to set the routing_mode path if it's not explicitly define. so due to that adding the path routing mode as the default one in 9.2+ installation in cli


whitfiea:
@dixitgsathwara is this still required? there are conflicts that need resolving?

PR #2328, opened 2026-05-28
[patch] Add support for ALM Agent operator logs

mahhum:
Updating the mg-collect-mas-core to also pickup operator logs from the core namespace. 

### Testing

I ran the mustgather against our dev cluster and was able to successfully execute against updated change:

image


terc1997 (approved):
LGTM

PR #2329, opened 2026-05-29
[patch] apply pre-install RBAC for 92FC to 920 update and fix for upgrade pipeline

Aneri11U:
## Description
- Updated upgrade and update flow similar to install to check user permissions before applying RBAC. Made changes in update flow to handle admin-mode for GA transition update. 
- This PR introduces a centralized [permission denial handler](https://github.com/ibm-mas/cli/blob/upgrade-fix/python/src/mas/cli/rbac_utils.py#L23C5-L23C34) to reduce the duplication of permission errors messages across workflows
- Updated install to use the permission denial handler. 

**Update**:
- The update workflow now detects when MAS instances are transitioning from pre-release to GA versions and automatically applies the required pre-install RBAC. It handles multiple instances, detects the appropriate admin mode for each, and provides clear guidance when user permissions are insufficient.

Update flow:

Initial Setup & Catalog Selection

RBAC Evaluation Phase

For each MAS instance in the cluster:
- Checks if current version is pre-release (e.g., 9.2.0-pre.stable+21734)
- Retrieves target version from catalog based on instance's channel
- shouldApplyRBACForInstance() determines if RBAC is needed:
Current = pre-release AND Target = GA AND Target >= 9.2.0 → RBAC Required
Current = GA OR Target = pre-release OR Target < 9.2.0 → Skip RBAC

For each instances needing RBAC:
- Detects admin mode (defaults to cluster for 9.2.0, detects existing for 9.3.0+)
- Retrieves installed apps using getInstalledAppsForRBAC()
- Stores instance metadata (ID, versions, channel, admin mode, apps)
- Filters out minimal mode instances (don't need pre-install RBAC)

Then checks if user has cluster-admin permissions If permissions exist then Sets applyPreInstallMASRBAC = True, if no permissions then displays warning with specific mas pre-install commands for each instance
Asks user to confirm RBAC was already applied by cluster administrator (and aborts if not)
And in last pipeline preparation phase - Applies Pre-Install RBAC if applyPreInstallMASRBAC set as true. 

**Upgrade**:
- Removed blocking upgrades when minimal admin mode detected. Applies RBAC only when user has permissions and mode is not minimal - If user doesn't have enough permissions then displays warning with specific mas pre-install command and ask user to confirm if that's already applied by the cluster administrator (and aborts if not)

**AI service** 
- Doesn't require specific RBAC to be applied for aiservice. 

## Jira Issue
[MASCORE-12627](https://jsw.ibm.com/browse/MASCORE-12627)
[MASCORE-13445](https://jsw.ibm.com/browse/MASCORE-13445)

## Testing
Tested on quick burn cluster
**Update** (for testing 9.2FC to 9.2.0  added [v9-000000-test-pre-dev.yaml](https://github.com/ibm-mas/python-devops/blob/8c09807d6df5095b5d9d3e41dc7bdbdbe0ffc288/src/mas/devops/data/catalogs/v9-000000-test-predev-amd64.yaml))
1. User who has access to apply pre-install RBAC
image
2. User who doesn't have access to apply pre-install RBAC

i. with --no-confirm flag
image
ii. without --no-confirm flag
image

**Upgrade** (for testing added the upgrade path from 9.2.x-feature to 9.2.x-dev)
1. User who has access to apply pre-install RBAC
image
2. User who doesn't have access to apply pre-install RBAC

i. with --no-confirm flag
image
ii. without --no-confirm flag
image
For 9.2+ upgrades, removed blocking for minimal mode
image
For other two modes - it will detect and apply RBAC according (if have permission to do so)
image
image

## Guidance for Reviewers
Before merging this PR, this PR must be merged first: https://github.com/ibm-mas/python-devops/pull/364


IanBoden (dismissed):
We need some unit tests for the new functions like evaluatePreInstallRBACAccess, shouldApplyRBACForInstance and handle_rbac_permission_denied.

Do we need 2 different evaluatePreInstallRBACAccess or should it be one common function? the mas variant loops through all mas instances on a cluster but the aiservice one doesn't, is that because we only support a single aiservice instance per cluster? The mas one also has code for defaulting to cluster when transitioning to 9.2.0 don't we also need that for aiservice?

Aneri11U:
> We need some unit tests for the new functions like evaluatePreInstallRBACAccess, shouldApplyRBACForInstance and handle_rbac_permission_denied.
> 
> Do we need 2 different evaluatePreInstallRBACAccess or should it be one common function? the mas variant loops through all mas instances on a cluster but the aiservice one doesn't, is that because we only support a single aiservice instance per cluster? The mas one also has code for defaulting to cluster when transitioning to 9.2.0 don't we also need that for aiservice?

For aiservice, I checked on aiservice [directory](https://github.com/ibm-mas/cli/tree/master/python/src/mas/cli/aiservice) only and saw install and upgrade :sweat_smile: Completely missed the update for aiservice. Thanks for pointing this out. Will make changes for considering aiservice during update. Yes I will add code for defaulting to cluster when transitioning to 9.2.0 for aiservice too.

IanBoden (approved):

PR #2337, opened 2026-06-03
[minor] Support June Catalog Update

prveenkumr:
June catalog includes:

MAS 9.2.0 GA
Patch updates for all in-support releases
DB2U v12

- https://github.ibm.com/maximoappsuite/ibm-maximo-operator-catalog/pull/1037
- https://github.com/ibm-mas/python-devops/pull/357
- https://github.com/ibm-mas/ansible-devops/pull/2278
- https://jsw.ibm.com/browse/MASCORE-14632


python/src/mas/cli/cli.py
```diff
@@ -136,7 +136,12 @@ def __init__(self) -> None:
         self.installRBACDir: str = path.join(self.cliRootDir, "rbac", "install")
         if not path.isfile(path.join(self.installRBACDir, "kustomization.yaml")):
             logger.error(f"Could not find install RBAC bundle at {self.installRBACDir}, this is required for the setup-rbac command")
-            self.installRBACDir = "/opt/app-root/src/rbac/install"
+            # Try alternative path for containerized environment
```
durera: What's the background behind this?  We've been testing this without this change for a number of months now I was under that impression anyway.  What has changed in the monthly update branch versus master that this is needed?

durera: It needs fixing anyway, because we can't hardcode paths like this ... everything needed for the install needs to be packaged within the python module, and accessed relative to where the module is installed, otherwise it won't work outside of a containerized environment - e.g. with uvx

python/src/mas/cli/cli.py
```diff
@@ -232,10 +242,12 @@ def __init__(self) -> None:
...
+            "9.2.x": " - <u>https://ibm.biz/MAS92-License</u>\n - <u>https://ibm.biz/MAXIT92-License</u>\n - <u>https://ibm.biz/MAXESRI92-License</u>\n\nBe aware, this channel subscription is supported for non-production use only.   \nIt allows early access to new features for evaluation in non-production environments.   \nThis subscription is offered alongside and in parallel with our normal maintained streams.   \nWhen using this subscription, IBM Support will only accept cases for the latest available bundle deployed in a non-production environment.   \nSeverity must be either 3 or 4 and cases cannot be escalated.   \nPlease refer to IBM documentation for more details.\n",
```
durera: looks like copied the 9.2.x-feature text

prveenkumr: fixed

python/src/mas/cli/cli.py
prveenkumr: Improved RBAC path resolution in [python/src/mas/cli/cli.py:139-144](vscode-webview://0igcs2qsscguftdbuqr281l4abiroekju3guqfkch491jsv7cfk3/python/src/mas/cli/cli.py:139-144)

Added fallback logic to check alternative path before using hardcoded fallback
Now tries the path relative to the package location before falling back to /opt/app-root/src/rbac/install
This addresses the error about not finding the RBAC bundle at the expected location

durera: My question is more about why is going up 5 directory levels from this file the right answer here?

`../../../../rbac/install`

That could take us anywhere if I'm running an install with `uvx mas-cli` command.

The hardcoded `self.installRBACDir = "/opt/app-root/src/rbac/install"` definitely isn't right, because it only works in the context of the container image .. however this fix only makes sense when running the command locally in the repo because `../../../../` will take you back to the project root .. that doesn't fix the underlying problem that the rbac needs to be packaged inside the python package, and the path to it needs to be relative to the package root dir.  

e.g.

```
        self.cliRootDir: str = path.join(path.abspath(path.dirname(__file__)), "..", "..", "..", "..")  # Bad - nothing should be using "cliRootDir"
        self.templatesDir: str = path.join(path.abspath(path.dirname(__file__)), "templates")  # Good - directory is inside the package
```

To make the rbac portable they need to be included in the package the same way the templates are:

```
[tool.setuptools.package-data]
"mas.cli" = [
  "templates/*.yaml",
  "templates/*.yml.j2",
  "templates/pod-templates/best-effort/*.yml",
  "templates/pod-templates/guaranteed/*.yml",
  "templates/pod-templates/saas-essentials/*.yml",
]
```

This way when the package is installed, those files are part of what is installed, and can be accessed from inside the `python/lib/site_packages/mas/cli` directory ... the code as it works now is going to be looking for `site_packages/rbac/install`


Look how `self.templatesDir` is being set up:

```
self.templatesDir: str = path.join(path.abspath(path.dirname(__file__)), "templates")
```

That gives us a directory within the package where we can access files that are part of the package.  That's what's needed here, otherwise this is broken anywhere outside of use in the container.

durera (approved):

python/src/mas/cli/cli.py
durera: As discussed, we will address this in a separate PR

PR #2346, opened 2026-06-04
[patch] Fix Monitor FVT to wait for IoT installation completion

jvaneel-dwivedi:
Changes waitfor-iot when condition to trigger when either launchfvt_iot OR launchfvt_monitor is true. This ensures IoT installation completes before Monitor FVT starts, even when IoT FVT is disabled.

The fix maintains David's simplified post-refactor logic that works for both pre-9.2 and post-9.2:
- waitfor-iot and waitfor-monitor run in parallel after approval-manage
- launchfvt-iot waits for both approval-iot AND approval-monitor
- launchfvt-monitor waits for launchfvt-iot (if enabled) OR approval-iot (if IoT FVT disabled)

Fixes issue in FVT CPD scenario where launchfvt_iot=false and launchfvt_monitor=true: Previously, waitfor-iot would not run, causing Monitor FVT to start before IoT installation completed. Now, waitfor-iot runs whenever Monitor FVT is enabled, ensuring proper sequencing.


durera (approved):
Looks good .. logic is now "wait for iot if testing iot or monitor" and "wait for monitor if testing iot or monitor" ... 

Note: I think it would impact the edge case of testing iot only (91 and earlier) or monitor only (9.2 and later), but I assume the monitor team are okay with that.

PR #2359, opened 2026-06-08
[patch] Allow arcgis installation on when admin-mode is cluster

SahilSheikh1:
ArcGIS installation is now allowed only when admin-mode is set to cluster.

I tested the installation with admin-mode=namespaced and admin-mode=minimal, and in both cases it failed with the expected fatal error.

**With admin-mode=minimal:**
image

**With admin-mode=namespaced:**
image

**With admin-mode=cluster:**
image


terc1997 (commented):
LGTM

IanBoden (approved):

PR #2433, opened 2026-06-26
[minor] Avoid generating DRO GitOps resources when DRO specs are not configured

nikitakumble:
### issue
https://jsw.ibm.com/browse/MASCORE-10911

### Description
This PR updates the provisioner logic to skip generating DRO-related GitOps resources when DRO specifications are not defined in `cluster-params`.

Previously, DRO files were generated even for MCSP provisioner clusters where no DRO specs were configured. With this change, DRO resources are generated only when the corresponding DRO configuration is present.

### Changes
Added a `dro_enabled` parameter to the Tekton pipeline.
Set `dro_enabled` based on the presence of the `ibm_dro` configuration in `cluster-params.yaml` from `saas-envs`.

### Testing
Tested both scenarios: verified that `gitops-dro.yaml` is generated only when `ibm_dro` is defined in `cluster-params.yaml` and is skipped otherwise.

### Related PR
https://github.ibm.com/maximoappsuite/saas-tekton/pull/258


mnivedithaa (dismissed):

tekton/src/pipelines/gitops/gitops-mas-cluster.yml.j2
```diff
@@ -468,6 +471,10 @@ spec:
     - name: gitops-dro
       runAfter:
         - gitops-cluster
+      when:
+        - input: "$(params.dro_enabled)"
```
whitfiea, line 475: for the other tasks we use install action and a value of install. For this dro_enabled, if I added `dro_enabled: false` then it should be installed.

whitfiea (changes requested):

tekton/src/pipelines/gitops/gitops-mas-cluster.yml.j2
nikitakumble: `dro_enabled` is not a user input, it's an internally derived parameter.
The requirement is to skip the `gitops-dro` task when the `ibm_dro` section is absent from `cluster-params.yaml`. To achieve this, `dro_enabled` is set from `$(event.merged_params.ibm_dro)`.
If `ibm_dro` exists, `dro_enabled` contains the object (non-empty), so the task runs.
If `ibm_dro` is removed, `dro_enabled` resolves to `""`, so the task is skipped.
Therefore, checking `notin [""]` correctly determines whether the `ibm_dro` configuration is present.

I agree the name `dro_enabled` is a bit misleading. Something like `is_dro_present` would better reflect its purpose.

whitfiea (dismissed):

whitfiea (approved):

PR #2441, opened 2026-06-29
[patch] removed perl dependency from facilities db2 setup script

dushyantbangal:
## Description

The perl script was being used to fetch the user's home directory. This is the only one perl dependency, and perl does not come as part of the db2 image. This required installing perl just for something so small.
This change uses shell script to achieve the same without using perl.
Updated the script with correct db2 ports.

## Testing
1. Downloaded db2 image `icr.io/db2_community/db2:11.5.9.0` (that is the version the script is based on)
2. Ran it using docker
3. Executed all the scripts
4. Checked connectivity to the database


terc1997 (approved):
LGTM

PR #2462, opened 2026-07-09
[patch] Add rhoai clusterrole for update SCC

karol-czarnecki:
Description:
We are trying to install Red Hat OpenShift AI (RHOAI) 2.25.8 on an OCP 4.2x cluster alongside our other applications.
During deployment we noticed that RHOAI creates a Security Context Constraint named openshift-ai-llminferenceservice-multi-node-scc.

Issue: 
two new created SCC with default changes break permissions for PVC in our pipeline
- openshift-ai-llminferenceservice-multi-node-scc
- openshift-ai-llminferenceservice-scc

Solution: 
Per comments here: https://ibm-cloudplatform.slack.com/archives/C067P31TWQ3/p1782745156816239 we can do workaround as fix from RHOAI is not expected for v2 - we are not sure if for v3. However we need to:
1. patch existing SCC with this: `opendatahub.io/managed: "false"` to annotation and labels 
2. patch existing SCC with this: `priority: null`

above changes are in ansible-devops


whitfiea (dismissed):
Can you update the PR description to state why this is needed as we normally don't what to create clusterroles

karol-czarnecki:
> Can you update the PR description to state why this is needed as we normally don't what to create clusterroles

Hi @whitfiea description added

rbac/install/pipeline/redhat-ods-operator.yaml
```diff
@@ -0,0 +1,47 @@
...
+  - kind: ServiceAccount
+    name: mas-{{ mas_instance_id }}-install-pipeline
+    namespace: mas-{{ mas_instance_id }}-pipelines
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: mas:{{ mas_instance_id }}:install-pipeline:redhat-ods-operator
+rules:
+  # Operator install
```
durera: Why do we need this permission for CRDs?

rbac/install/pipeline/redhat-ods-operator.yaml
```diff
@@ -0,0 +1,47 @@
...
+    resourceNames:
+      - openshift-ai-llminferenceservice-multi-node-scc
+      - openshift-ai-llminferenceservice-scc
+    verbs:
+      - list
+      - create
+      - get
+      - patch
+      - update
+  - apiGroups:
```
durera: Similar Q here, I thought the only extra we needed for this workaround was the modification of an existing SCC?

rbac/install/pipeline/redhat-ods-operator.yaml
```diff
@@ -0,0 +1,47 @@
...
+      - patch
+  # RHOAI requires scc patch for workaround to not update PVC settings by operator
+  - apiGroups:
+      - security.openshift.io
+    resources:
+      - securitycontextconstraints
+    resourceNames:
+      - openshift-ai-llminferenceservice-multi-node-scc
+      - openshift-ai-llminferenceservice-scc
+    verbs:
```
durera, line 28: I don't think we need create do we?  We're only updating an existing one to fix it because Red Hat won't deliver a fix?

durera (changes requested):

rbac/install/pipeline/redhat-ods-operator.yaml
karol-czarnecki: removed this part

rbac/install/pipeline/redhat-ods-operator.yaml
karol-czarnecki: removed create from permission

rbac/install/pipeline/redhat-ods-operator.yaml
karol-czarnecki: removed this part

durera (approved):

PR #2501, opened 2026-07-31
[patch] Add interactive ODH/RHOAI platform selection prompt for AI Service installation

Divyesh-Khokhar:
## Summary

Adds an interactive prompt to the AI Service installation flow asking users to choose between **Red Hat OpenShift AI (RHOAI)** and **Open Data Hub (ODH)** as the AI data science platform.

Previously, interactive mode always defaulted to ODH silently. Non-interactive mode already supported \`--rhoai\` flag.

## Testing Snapshots
- Standalone AIService pipeline:
Screenshot 2026-07-31 at 12 18 30 PM

- Mas install pipeline:
Screenshot 2026-07-31 at 3 46 21 PM


karol-czarnecki (approved):
lgtm - however we should switch completely to use rhoai for 9.1.x. This solution is temporary once we stop use odh for 9.1.x

whitfiea (approved):

PR #2518, opened 2026-08-07
[patch] reenable routingMode

aadawadk:
allow routingMode to be set
https://dashboard.ibmmas.com/tests/aadawadk/8125

[#8167](https://dashboard.ibmmas.com/tests/aadawadk/8167)


python/tests/integration/install/test_dev_mode.py
```diff
@@ -505,7 +503,7 @@ def test_install_master_dev_mode_non_interactive(tmpdir):
     run_install_test(tmpdir, config)
 
 
-def test_install_master_dev_mode_non_interactive_with_path_routing(tmpdir, caplog):
```
durera, line 508: Is the removal of caplog intentional, what does it do/how does it relate to re-enabling routing mode??

aadawadk, line 508: Yes
It is intentional.
1. I reverted the changes from this PR 
https://github.com/ibm-mas/cli/pull/2391/changes
caplog was only used to capture and verify the "Path based routing mode not supported" error
2. Now that PBR is re-enabled, the test expects the install to succeed, so the error assertion and `caplog` are no longer needed.

unnati-solanki-git (approved):

PR #2537, opened 2026-08-17
[patch] CP4D 5.3.1 Helm airgapped install support

nehasangwai-ibm:
## Summary

Adds airgap (mirror) support for Helm charts shipped with Cloud Pak for Data 5.3.1 CASE bundles. This change 
ensures those charts are mirrored to/from the target registry as part of the standard 
`mas mirror` workflow.

### `python/src/mas/cli/mirror/config.py`
- Extended `PACKAGE_CONFIGS` tuples with a `has_HelmCharts` boolean flag to identify CPD CASE 
  bundles that also ship Helm charts (`ibm-cp-datacore`, `ibm-ccs`, `ibm-datarefinery`, 
  `ibm-wsl`, `ibm-wsl-runtimes`, `ibm-wml-cpd`, `ibm-redis-cp`, `ibm-analyticsengine`, 
  `ibm-cognos-analytics-prod`)
- Added `HELM_ONLY_CHART_CONFIGS` for bundles that have Helm charts but no standalone ISC 
  image files (`ibm-opencontent-opensearch`)
- Added `ibm-redis-cp` (WML dependency) to `PACKAGE_CONFIGS`

### `python/src/mas/cli/mirror/app.py`
- Added `mirrorCharts()` function supporting all three mirror modes:
  - **m2m**: pulls each chart into a temp dir then pushes to target OCI registry
  - **m2d**: pulls each chart to `<rootDir>/charts/<caseName>/<version>/`
  - **d2m**: pushes `.tgz` archives from disk to target OCI registry
- Added `getChartMetadata()` to fetch per-CASE chart metadata YAML from `ibm-mas/image-set-configs`
- Added `ChartMirrorResult` dataclass to track per-CASE chart mirror results
- Added `_helmLogin()` helper for OCI registry authentication
- Chart mirroring is gated on `cpd_product_version_default >= 5.2.0` from the catalog
- Mirror result summary now reports Helm chart failures separately from image failures

### `python/src/mas/cli/mirror/argParser.py`
- Minor fix

### `tests/integration/utils/mirror_test_helper.py`
- Added `redis_version` to the test helper mirror config

### Snippets

m2m mode: 
image

m2d mode: 
image

d2m mode: 
image


python/src/mas/cli/mirror/app.py
```diff
@@ -596,6 +598,280 @@ def mirrorCatalog(
...
+
+    Args:
+        registry: Registry hostname (and optional port), e.g. "myregistry:5000"
+        username: Registry username
+        password: Registry password
+
+    Returns:
+        True on success, False on failure
+    """
+    cmd = ["helm", "registry", "login", registry, "--username", username, "--password", password, "--insecure"]
```
durera: Similar comment to ansible-devops ... `--insecure` prob shouldn't be the default here, but something that is an option.

durera (changes requested):

python/src/mas/cli/mirror/app.py
SahilSheikh1: Pushed changes,
playbook will be reading "HELM_REGISTRY_CA_FILE" env. If set then it will pass it accordingly.

durera (approved):

PR #2552, opened 2026-08-24
[minor] Default to RHOAI and remove ODH/RHOAI interactive prompt

Divyesh-Khokhar:
## Summary

Removes the interactive "AI Data Science Platform" selection prompt from both `mas install` and `mas aiservice install`. Red Hat OpenShift AI(RHOAI) is now the default and only supported AI data science platform going forward.

## Changes

- **`mas install`** (`install/app.py`):
  - `configOperationMode()` now defaults `rhoai` param to `"true"`.
  - Removed the interactive ODH/RHOAI selection dialog in
    `aiServiceSettings()`; replaced with a warning notice.

- **`mas aiservice install`** (`aiservice/install/app.py`):
  - Non-interactive `non_prod` handler now sets `rhoai="true"` instead
    of `"false"` (RHOAI is now the default without needing `--rhoai`).
  - Removed the same interactive ODH/RHOAI selection dialog; replaced
    with the same warning notice.

## Testing
Screenshot 2026-08-24 at 2 57 42 PM
Screenshot 2026-08-24 at 3 27 13 PM

### DIFF
--- a/python/src/mas/cli/aiservice/install/app.py
+++ b/python/src/mas/cli/aiservice/install/app.py
@@ -313,14 +313,14 @@ class AiServiceInstallApp(BaseApp, aiServiceInstallArgBuilderMixin, aiServiceIns
                     self.setParam("environment_type", "production")
                     self.setParam("aiservice_odh_model_deployment_type", "raw")
                     self.setParam("aiservice_rhoai_model_deployment_type", "raw")
-                    self.setParam("rhoai", "false")
+                    self.setParam("rhoai", "true")
                 else:
                     self.operationalMode = 2
                     self.setParam("mas_annotations", "mas.ibm.com/operationalMode=nonproduction")
                     self.setParam("environment_type", "non-production")
                     self.setParam("aiservice_odh_model_deployment_type", "serverless")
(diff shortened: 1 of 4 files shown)


python/src/mas/cli/aiservice/install/app.py
```diff
@@ -313,14 +313,14 @@ def nonInteractiveMode(self) -> None:
                     self.setParam("environment_type", "production")
                     self.setParam("aiservice_odh_model_deployment_type", "raw")
                     self.setParam("aiservice_rhoai_model_deployment_type", "raw")
-                    self.setParam("rhoai", "false")
+                    self.setParam("rhoai", "true")
```
jasmin-30, line 316: Since RHOAI is now the default platform, should we remove the `rhoai` flag from the install pipeline altogether?

Divyesh-Khokhar: We were thinking of removing the `rhoai` flag once we completely remove the ODH task. At that point, we can remove the `rhoai` flag from the install pipeline as well.

jasmin-30 (approved):

whitfiea (approved):

PR #2559, opened 2026-08-25
[patch] Allow setting up routingMode from cli for Sept release

aadawadk:


IanBoden (changes requested):
As this stands we can't merge this till path based routing works and we are ready to release

PR #2567, opened 2026-08-28
[patch] added smartcleanse to AIService FVT pipeline

yohlema:
##Description
added smartcleanse to AIService FVT pipeline

##Test Results
image

### DIFF
--- a/tekton/src/pipelines/aiservice-fvt.yml.j2
+++ b/tekton/src/pipelines/aiservice-fvt.yml.j2
@@ -129,7 +129,7 @@ spec:
         - name: fvt_test_suite
           value: smartcleanse
       runAfter:
-        - fvt-aiservice-tenant-entitlement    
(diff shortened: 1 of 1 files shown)


tekton/src/pipelines/aiservice-fvt.yml.j2
```diff
@@ -120,7 +120,17 @@ spec:
           value: docsearch
       runAfter:
         - fvt-aiservice-tenant-entitlement
-    # 6. PCC test suite
+    # 6. SMARTCLEANSE test suite
```
jasmin-30, line 123: There are already 4 model tests being executed at simultaneously with smartcleanse. Can we update the sequence so that it executes along with nl2oslc-gpt model.

jasmin-30 (commented):
Values for runAfter clause for subsequent tasks need to be updated with newly added tasks.

